PR #9, opened 2019-09-13
Handle nested providers

brentvatne:
## Summary

People may accidentally nest SafeAreaProviders by using multiple libraries that both depend on it. While this is not great, we can handle it gracefully and let people use linters or something like that to avoid this instead.

## Test Plan

Works in this app:

```tsx
import React from 'react';
import {View, Text} from 'react-native';

import {SafeAreaProvider, useSafeArea} from 'react-native-safe-area-context';

function AppContainer() {
  return (
    <SafeAreaProvider>
      <App />
    </SafeAreaProvider>
  );
}

function App() {
  let insets = useSafeArea();

  return (
    <View style={{flex: 1, paddingTop: insets.top, alignItems: 'center'}}>
      <Text style={{fontSize: 30}}>Hello there</Text>
      <SafeAreaProvider>
        <Text style={{fontSize: 30}}>wat</Text>
      </SafeAreaProvider>
      <Text style={{fontSize: 30}}>ok</Text>
    </View>
  );
}

export default AppContainer;
```

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { StyleSheet } from 'react-native';
+import { StyleSheet, View } from 'react-native';
 import { EdgeInsets, InsetChangedEvent } from './SafeArea.types';
 import NativeSafeAreaView from './NativeSafeAreaView';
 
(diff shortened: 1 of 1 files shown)


a-eid:
should one use `SafeAreaProvider` on native ios/android? 
what does it do exactly ?

brentvatne:
it renders a view that determines safe area information and then passes it down through context

Issue #10, opened 2019-09-30
Suggestion - Update Readme with more information


rtman:
Hi,

I'm interested in using this package, but I'm a little wary of implementing it since the readme is kind of light on details. 

For instance, does this work with all android and ios devices with notches? 
What does this package actually do? (description of how it works)
Will this be actively maintained?
How does this handle devices without notches?

If you could add a bit more to the readme I think that would help people to decide on using this. It's coming recommended from @mikehardy via `react-native-device-info` as a replacement for their `hasNotch` method, so maybe adding that could also help?

mikehardy:
@janicduplessis this is spilling over from discussion on react-native-device-info

@rtman here is the previous thread: https://github.com/react-native-community/react-native-device-info/issues/594#issuecomment-527669853

It does not exist yet but we are thinking it is possible to expose it so then you can answer the boolean "hasNotch" question

mikehardy:
So - a follow on thought @rtman if you are doing your own padding and don't want some other package just auto-fixing it then until some new API existed then it wouldn't work no - for that reason it is a conversation at the moment, not a formal API move and breaking change over on RNDI etc

rtman:
@mikehardy So the only reason we add padding to non notched devices is because react-native's `SafeAreaView` adds vertical padding for ios notched devices. We don't want to add more vertical padding so we just removed it entirely and only add vertical padding to non notched devices. If this package can handle that, then I wouldn't need `hasNotch`.

However since this isn't implemented yet (as far as I know), I'll need to continue using `hasNotch` with `react-native-device-info`. Which is currently bugged by the sync + chrom debugger issue. A constant to check in the meantime would be much appreciated!

janicduplessis:
@rtman The concept of safe area is a bit different than just if the device has a notch. For example the area under the status bar is considered in the safe area zone even when the device has no notch.

I think what you could do in your case is just use the maximum between some static value and the safe area value. So let's say you want to add padding at the bottom of a view of at least 20px you could do `<View style={{ paddingBottom: Math.max(safeAreaInsets.bottom, 20) }} />`.

janicduplessis:
Going to close this, if anyone (or myself when I get time) wants to add more details to the readme feel free to open a PR.

Issue #26, opened 2019-11-20
Export initial mount window insets


jacobp100:
If we exported the insets of the window when the app launched, we could avoid delaying the rendering of the subtree, which could improve initial mount performance

```jsx
import {SafeAreaProvider, initialMountWindowInsets} from 'react-native-safe-area-context';

<SafeAreaProvider initialSafeAreaInsets={initialMountWindowInsets}>
  {content}
</SafeAreaProvider>
```

We might even be able to make this the default. But I'm not sure that's safe (and I'm not sure the current behaviour of inheriting the parent's insets is safe either) because if you present a modal in page sheet mode, the insets will be different from the root

We'd need to check the value of the insets when they get updated from the `onInsetsChange` event to avoid reconciling the subtree twice

Is this a change you'd accept a PR for?

janicduplessis:
Added android support too, let's close this!

PR #27, opened 2019-11-23
Add support for getting initial safe area insets on mount

jacobp100:
## Summary

Adds support for getting initial safe area insets on mount. See #26 

## Test Plan

`console.log` of the value on verified this value was correct

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,27 @@ function SomeComponent() {
 
 If you are doing server side rendering on the web you can use `initialSafeAreaInsets` to inject insets value based on the device the user has, or simply pass zero values. Since insets measurement is async it will break rendering your page content otherwise.
 
+### Optimization
+
+To speed up the initial render, you can import `initialSafeAreaInsets` from this package and set in on the provider as described in Web SSR. You cannot do this if your provider remounts, or you are using `react-native-navigation`.
+
+Only supported on iOS at the moment.
+
+```js
+import {
+  SafeAreaProvider,
+  initialSafeAreaInsets,
+} from 'react-native-safe-area-context';
+
+function App() {
+  return (
+    <SafeAreaProvider initialSafeAreaInsets={initialSafeAreaInsets}>
+      ...
+    </SafeAreaProvider>
+  );
+}
+```
+
 ## Resources
 
 - Great article about how this library can be used: https://dev.to/brunolemos/adding-notch-support-to-your-react-native-android-app-3ci3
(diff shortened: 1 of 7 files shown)


janicduplessis:
@jacobp100 Thanks for working on this! Could you rename this to `initialWindowInsetsSafeAreaInsets` so it's clear that it is insets from the root window and that they can be different from the one returned by the provider (in the case it doesn't cover the screen).

Also what do you think about exposing this as a prop on the provider vs this api?

Something like `initializeWithWindowInsets` maybe? Might be safer so people don't rely on values provided by `initialWindowInsetsSafeAreaInsets` for other things.

satya164:
How about `initialWindowSafeAreaInsets` instead of `initialWindowInsetsSafeAreaInsets`?

> Also what do you think about exposing this as a prop on the provider vs this api?

Does a provider have any advantage instead of an exported constant? I guess this value is not really constant and could change on device orinetation change etc.?

janicduplessis:
@satya164 Oups that was a typo lol, I'll edit my comment.

It would allow not exposing the actual value so people don't use it because it is convenient, since like you said it can change. The provider could use the value internally to initialize itself though. It would still have to be disabled by default since we can't assume the provider covers the screen (for example if you have a provider inside the new ios modal).

jacobp100:
> Could you rename this to initialWindowSafeAreaInsets so it's clear that it is insets

Will do

To me, adding another prop for this feels like increasing the API surface area just so people don't make these incorrect assumptions. We might be better documenting all the cases that affect the safe area insets (also fixes #10), and saying out that it's never safe to rely on this value past the initial mount

Or half way would be we expose it as an opaque object (i.e. `{ __preset: "window" }`) so the user couldn't rely on it

Or we can just do the prop - it's all subjective anyway! Lemmi know what way you wanna go

We should do the docs anyway. I'm happy to do iOS docs. I can take a look at Android and make a best attempt, but I'm pretty unfamiliar with the notch situation there, so might need review

janicduplessis:
Ok let's just rename it to `initialWindowSafeAreaInsets` and go with that. It might be possible to implement on Android using something like `activity.getWindow().getDecorView().getRootWindowInsets()` but we can do that in a follow up.

jacobp100:
Hey sorry for the delay - I didn't get the notification from Github. I've done this change now. I'll try and and look at Android at some point in a future PR

janicduplessis:
@jacobp100 Thanks for this!

PR #33, opened 2019-12-14
fix: Android stand-alone build.gradle with .gitignore

Rapsssito:
## Summary

Added `buildscript` and `repositories` attributes to `buid.gradle`, required when opening the android folder stand-alone. This avoids unnecessary downloads and potential conflicts when the library is included as a module dependency in an application project.


janicduplessis:
We're already using `files` in package.json, I don't think the npmignore is needed.

Rapsssito:
@janicduplessis I didn't see it sorry. I removed `.npmignore`.

janicduplessis:
Looks good, thanks!

PR #34, opened 2019-12-14
fix: Android return dp instead of px

Rapsssito:
## Summary

React Native uses density-independent pixels, or **dp**, as it's default unit. This will size elements so that they are roughly the same physical size on different devices. However, on Android we are returning pixels (px) instead of dp, resulting in a non-consistent look on different devices.

### DIFF
--- a/android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaView.java
+++ b/android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaView.java
@@ -44,14 +44,14 @@ public class SafeAreaView extends ReactViewGroup implements ViewTreeObserver.OnG
     } else {
       int rotation = mWindowManager.getDefaultDisplay().getRotation();
       float statusBarHeight = 0;
-      int resourceId = resources.getIdentifier("status_bar_height", "dimen", "android");
+      int resourceId = getResources().getIdentifier("status_bar_height", "dimen", "android");
       if (resourceId > 0) {
-        statusBarHeight = PixelUtil.toDIPFromPixel(resources.getDimensionPixelSize(resourceId));
+        statusBarHeight = PixelUtil.toDIPFromPixel(getResources().getDimensionPixelSize(resourceId));
       }
       float navbarHeight = 0;
-      resourceId = resources.getIdentifier("navigation_bar_height", "dimen", "android");
+      resourceId = getResources().getIdentifier("navigation_bar_height", "dimen", "android");
       if (resourceId > 0) {
-        navbarHeight = PixelUtil.toDIPFromPixel(resources.getDimensionPixelSize(resourceId));
+        navbarHeight = PixelUtil.toDIPFromPixel(getResources().getDimensionPixelSize(resourceId));
       }
 
       windowInsets = new EdgeInsets(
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaView.java
```diff
@@ -45,12 +45,12 @@ private EdgeInsets getSafeAreaInsets() {
       int statusBarHeight = 0;
       int resourceId = getResources().getIdentifier("status_bar_height", "dimen", "android");
       if (resourceId > 0) {
-        statusBarHeight = getResources().getDimensionPixelSize(resourceId);
+        statusBarHeight = (int) (getResources().getDimension(resourceId) / getResources().getDisplayMetrics().density);
```
janicduplessis: Can you use the helper from react native `PixelUtil.toDIPFromPixel`

Rapsssito: Successfully changed to `PixelUtil.toDIPFromPixel`

android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaView.java
```diff
@@ -45,12 +46,12 @@ private EdgeInsets getSafeAreaInsets() {
       int statusBarHeight = 0;
       int resourceId = getResources().getIdentifier("status_bar_height", "dimen", "android");
       if (resourceId > 0) {
-        statusBarHeight = getResources().getDimensionPixelSize(resourceId);
+        statusBarHeight = Math.round(PixelUtil.toDIPFromPixel(getResources().getDimensionPixelSize(resourceId)));
```
satya164: Is rounding necessary? If it's actually not a round value, wouldn't it cause incorrect height?

Rapsssito: You are right! Changed.

janicduplessis:
@Rapsssito I had another look and can you explain why the value needs to be converted to dp here? This code uses pixel values everywhere and is converted to dp before being sent to JS here https://github.com/th3rdwave/react-native-safe-area-context/blob/master/android/src/main/java/com/th3rdwave/safeareacontext/InsetsChangeEvent.java#L28. This code is only used to polyfill `getSystemWindowInset*` on old android versions. AFAIK `getSystemWindowInset*` returns pixel values.

Rapsssito:
> 
> 
> @Rapsssito I had another look and can you explain why the value needs to be converted to dp here? This code uses pixel values everywhere and is converted to dp before being sent to JS here https://github.com/th3rdwave/react-native-safe-area-context/blob/master/android/src/main/java/com/th3rdwave/safeareacontext/InsetsChangeEvent.java#L28. This code is only used to polyfill `getSystemWindowInset*` on old android versions. AFAIK `getSystemWindowInset*` returns pixel values.

I took another look, too. It turns out that the custom version of the library I am using does not use the class `InsetsChangeEvent`, so that is why I was receiving px instead of dp.

Sorry if I wasted your time!

janicduplessis:
@Rapsssito No problem! Thanks for trying to contribute back :)

PR #41, opened 2020-01-19
fix(build): npmignore unwanted platform-specific files

### LINKED ISSUE #24: Unexpected .iml file in npm package
The npm package has an unexpected file `android/react-native-safe-area-context.iml` which may cause
```
Could not set process working directory to '/node_modules/react-native-safe-area-context/example/android': could not setcwd() (errno 2: No such file or directory)
```
in Android Studio.

## Solution
add `*.iml` to `.npmignore`

mikehardy:
## Summary

Fix #24 by positively identifying directories + files we want in package.json files section, then using npmignore files to ignore unwanted platform files

## Test Plan

These is the same package.json::files +  .npmignore combo style we use in react-native-device-info to make better packages, courtesy of @Friederbluemle https://github.com/react-native-community/react-native-device-info/pull/872

```
(base) mike@kunashir:~/work/react-random/react-native-device-info (master) % cp ios/.npmignore ../react-native-safe-area-context/ios/
(base) mike@kunashir:~/work/react-random/react-native-device-info (master) % cp android/.npmignore ../react-native-safe-area-context/android/
```

### DIFF
--- a/android/.npmignore
+++ b/android/.npmignore
@@ -1,0 +1,10 @@
+*.iml
+.DS_Store
+.gradle/
+.idea/
+.npmignore
+build/
+gradle/
+gradlew
+gradlew.bat
+local.properties
--- a/ios/.npmignore
+++ b/ios/.npmignore
@@ -1,0 +1,6 @@
+*/project.xcworkspace/
+*/xcuserdata/
+.DS_Store
+.npmignore
+Pods/
+build/
--- a/package.json
+++ b/package.json
@@ -11,7 +11,6 @@
     "/src",
     "/lib",
     "/android",
-    "!/android/build",
     "/ios",
     "/*.podspec"
   ],


mikehardy:
@satya164 - from @friederbluemle on https://github.com/react-native-community/react-native-device-info/pull/872#issuecomment-548124928

> What I found was that with or without the leading slash, only root directories (basically on the same level as the package.json) are considered.
In other words: The entries in files don't match nested folders (a behavior that's different .gitignore).

satya164:
So this?

```js
/android/src
/android/build.gradle
/ios/SafeAreaView
/ios/SafeAreaView.xcodeproj/project.pbxproj
``

**Comment by mikehardy** [comment]
My understanding is leading slashes did not matter, and only top-level directories were considered, so no not that, but you will have to run `npm pack` and check the output to test it

**Comment by friederbluemle** [comment]
Yes, leading slashes are _not_ necessary in the `files` array. Only top level objects (from where `package.json` is located) are considered anyway.
The output of `npm pack --dry-run` will also list matching files _without_ the leading slash.
@satya164's list would also work (for this project), whereas the combination of the proposed files array together with the selective `.npmignore` files inside of `android/` and `ios/` is a generic solution that should work for almost every RN library project.

**Comment by satya164** [comment]
Just tested:

```js
~/Workspace/test
❯ tree
.
├── android
│   ├── build
│   ├── file.json
│   ├── src
│   │   └── main.xml
│   └── test.java
├── bar.js
├── index.js
└── package.json

2 directories, 7 files
~/Workspace/test
❯ npm pack
npm notice
npm notice 📦  test@1.0.0
npm notice === Tarball Contents ===
npm notice 0    index.js
npm notice 0    android/file.json
npm notice 162B package.json
npm notice 0    android/src/main.xml
npm notice === Tarball Details ===
npm notice name:          test
npm notice version:       1.0.0
npm notice filename:      test-1.0.0.tgz
npm notice package size:  272 B
npm notice unpacked size: 162 B
npm notice shasum:        f956f603dfbb4521527bdb3bf8c43128e274ff27
npm notice integrity:     sha512-iytim3p1eBzVj[...]T3sTEz6OG3e+w==
npm notice total files:   4
npm notice
test-1.0.0.tgz
~/Workspace/test
❯ tar -tf test-1.0.0.tgz
package/index.js
package/android/file.json
package/package.json
package/android/src/main.xml
```

satya164:
> the selective .npmignore files inside of android/ and ios/ is a generic solution that should work for almost every RN library project.

The `files` array should always be preferred not to commit unnecessary files to `npm` imo.

mikehardy:
If the smaller (files-only) method works for this repo, works for me, I just want #24 fixed as it seems to be driving my Android Studio crazy :-). Feel free to implement however and close this at will :-)

PR #43, opened 2020-01-20
fix: inset bottom with navbarHeight only on landscape

### LINKED ISSUE #39: Incorrect insets.bottom value on the devices with Android 5.x.x and hardware buttons
The package returns incorrect `insets.bottom` value on the devices with Android 5.x.x and hardware buttons. E.g. Samsung S4.

Code example:
```js
import { useSafeArea } from 'react-native-safe-area-context';

...

const Component = () => {
  const insets = useSafeArea();
  console.log(insets.bottom); // prints 48 - it should be 0

  ...
}
```

luancurti:
## Summary

Inset bottom is wrong using Android 5.x.x when is in Portrait

Fixes #39 

## Test Plan

Test on device with Android 5.x.x in Portrait and inset bottom should be 0


luancurti:
@janicduplessis The left and right insets is wrong, setting the insets 0 from `bottom, right and left` this is the result. In API LEVEL 23 or earlier has android devices with notched if not I think this is the better approch, what do you think?

Screenshot_1580417362

Screenshot_1580417369

Screenshot_1580417375

Screenshot_1580417379

janicduplessis:
I think you can have a transparent navigation bar before api 23 though, this code was to account for that if I remember correctly.

PR #70, opened 2020-05-04
Remove unnecessary use_native_modules!

luancurti:
## Summary

Fix warning `use_native_modules! skipped the react-native dependency @react-native-community/bob. No podspec file was found` when call `use_native_modules!` this warning appear after run `pod install` passing the `path` to the `use_react_native!` fixed this

## Test Plan

Run` pod install` and build the app with the `Xcode` or `react-native run-ios` and the app should `work`


janicduplessis:
👍 Thanks!!

PR #78, opened 2020-05-20
Native SafeAreaView

jacobp100:
## Summary

* Renames `RNCSafeAreaView` -> `RNCSafeAreaProvider`
* Implements native `SafeAreaView` (as `RNCSafeAreaView`) avoiding JS bridge delay (iOS only for now)
* Makes setting padding on `SafeAreaView` additive rather than replacing the safe area (breaking change)
* Allows customising which edges to apply padding to on `SafeAreaView`

Android implementation to come at a later date

## Test Plan

Verify `SafeAreaView` still works. Verify adding padding works additively (rather than replacing). Verify setting `edges` prop.


example/src/App.tsx
```diff
@@ -48,22 +51,19 @@ export default function App() {
...
+      return <SafeAreaViewExample />;
+    case 'hooks':
+      return <HooksExample />;
     case 'react-navigation-4':
       return <ReactNavigation4Example />;
     case 'react-navigation-5':
       return <ReactNavigation5Example />;
     case 'native-stack':
       return <NativeStackExample />;
-    case 'react-native-safe-area-view':
```
janicduplessis, line 65: Let's leave this example, I added to avoid breaking react-native-safe-area-view since it uses this library.

ios/SafeAreaView/RNCSafeAreaProviderManager.m
```diff
@@ -0,0 +1,47 @@
...
+        @"frame": @{
+          @"x": @(window.frame.origin.x),
+          @"y": @(window.frame.origin.y),
+          @"width": @(window.frame.size.width),
+          @"height": @(window.frame.size.height),
+        },
+      }
+    };
+  } else {
+    return @{ @"initialWindowSafeAreaInsets": [NSNull null] };
```
janicduplessis, line 43: I just realized this should be `initialWindowMetrics` and not `initialWindowSafeAreaInsets`, would you mind fixing it while here?

ios/SafeAreaView/RNCSafeAreaProvider.m
```diff
@@ -0,0 +1,67 @@
+// Simplified version of https://github.com/facebook/react-native/blob/master/React/Views/RNCSafeAreaView/RCTRNCSafeAreaView.m
```
janicduplessis, line 1: I think it was there before but can you remove this comment, it is not really relevent anymore.

ios/SafeAreaView/RNCSafeAreaViewEdges.m
```diff
@@ -0,0 +1,29 @@
+#import <React/RCTConvert.h>
+#import "RNCSafeAreaViewEdges.h"
+
+@implementation RCTConvert (RNCSafeAreaView)
+
+RCT_ENUM_CONVERTER(RNCSafeAreaViewEdges, (@{
```
janicduplessis, line 6: Not a fan of this API, what about using something like:

```ts
export type Edges = {
  top?: boolean;
  bottom?: boolean;
  left?: boolean;
  right?: boolean;
};

type Props = {
  disableEdges: Edges;
  // or
  edges: Edges;
};

<SafeAreaView disableEdges={{top: true}} />
// or
<SafeAreaView edges={{top: false}} />
```

janicduplessis: I don't mind keeping the bitmap as an implementation detail thought.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-native-safe-area-context",
-  "version": "1.0.2",
+  "version": "2.0.0",
```
janicduplessis, line 3: Leave the version like this I'll take care of updating that when doing a release.

jacobp100:
Thanks for taking the time to review! I’ll get the fixes in and take a stab at the android side in this PR

The only thing I’d say is an object for the edges will be more difficult to write and read - especially if people aren’t consistent with the order of edges. Not a deal breaker, but I’ll have a think if there’s any alternatives.

ios/SafeAreaView/RNCSafeAreaViewEdges.m
jacobp100: What do you think about `edges={['top', 'bottom']}`?

jacobp100:
I've made the changes in your comments - let me know if I missed anything.

I added an Android implementation too. I'm not too confident with Android/Java, so any pointers on stuff like style would be useful!

For the edges, I went with an array of edges - it's similar to the format of the `dataDetectorTypes` prop.

README.md
```diff
@@ -114,39 +152,27 @@ class ClassComponent extends React.Component {
...
-
-### Web SSR
+## Web SSR
 
 If you are doing server side rendering on the web you can use `initialMetrics` to inject insets and frame value based on the device the user has, or simply pass zero values. Since insets measurement is async it will break rendering your page content otherwise.
 
-### Optimization
+## Optimization
+
+If you can, use `SafeAreaView`. It's implemented natively on iOS so when rotating the device, there is no delay from the asynchronous bridge. A native Android implementation is upcoming.
```
janicduplessis, line 168:
```suggestion
If you can, use `SafeAreaView`. It's implemented natively so when rotating the device, there is no delay from the asynchronous bridge.
```

android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaView.java
```diff
@@ -1,40 +1,57 @@
 package com.th3rdwave.safeareacontext;
 
+import static com.facebook.react.uimanager.UIManagerHelper.getReactContext;
+
 import android.annotation.SuppressLint;
 import android.content.Context;
+import android.util.Log;
```
janicduplessis, line 7: Remove

jacobp100, line 7:
```suggestion

```

jacobp100:
Thanks 😄

janicduplessis:
Published as 2.0.0-beta.0

PR #80, opened 2020-05-22
Docs: linking in RN 0.60 is automatic

fson:
## Summary

Linking with `react-native link` is no longer necessary in RN >= 0.60.

Screen Shot 2020-05-22 at 17 16 06


janicduplessis:
Thanks!!

PR #50, opened 2020-02-06
Attempt to resolve #48

hyochan:
## Summary

Resolve #48

## Test Plan

After changing above and run build, I now see different error.
```
  ● Test suite failed to run

    Invariant Violation: __fbBatchedBridgeConfig is not set, cannot invoke native modules

      at invariant (node_modules/react-native-safe-area-context/node_modules/invariant/invariant.js:40:15)
      at Object.invariant (node_modules/react-native-safe-area-context/node_modules/react-native/Libraries/BatchedB
ridge/NativeModules.js:166:3)
      at Object.require (node_modules/react-native-safe-area-context/node_modules/react-native/Libraries/TurboModul
e/TurboModuleRegistry.js:13:23)

Test Suites: 1 failed, 5 passed, 6 total
```

PR #85, opened 2020-05-25
Margin mode

jacobp100:
#50  Summary

Adding margin mode to safe area view

I noticed while doing this that a view's safe area is impacted by the position of itself. If you set a margin, the safe area gets decreased, so you get an infinite loop.

To work around this, I use the parent view controller's view's safe area (i.e. a modal, stack navigation controller, or root view controller). I removed the logic from Android that mimicked iOS's behaviour so they're now hopefully the same.

I'm hoping this change will make it match closer to v1 SafeAreaView, but I'm not sure if this will cause other issues - what do you think?

Note - I looked briefly border modes too - but they use floats instead of yoga values, so made it much more complicated to implement

## Test Plan

We have a new playground

Simulator Screen Shot - iPhone 11 - 2020-05-26 at 18 36 00


janicduplessis:
I think it would be nice to make android apply insets the same way as ios when it is not touching edges. But that's for another time :)

Is this ready to review?

jacobp100:
Just pushed the Android implementation and updated the PR description

jacobp100:
If we go this route, I just realised the fallback on iOS 10 would need updating to just use the parent view controller's insets. Will update tomorrow

PR #86, opened 2020-05-25
Add missing UIKit import

tsapeta:
## Summary

I've been updating this library in Expo for SDK38 and then the build failed due to `Unknown type name 'UIEdgeInsets'` error.

## Test Plan

Expo client with vendored code from v2.0.1 including this change compiled successfully. It cannot break anything 😄


janicduplessis:
👍  Thanks!

Issue #31, opened 2019-12-09
how to properly mock this package?


sibelius:
Without mocking this package, react-navigation won't render any route in jest tests

janicduplessis:
I see, you can render a SafeAreaProvider with initialSafeAreaInsets at the top of your test view hierarchy. See https://github.com/th3rdwave/react-native-safe-area-context/blob/master/src/__tests__/index-test.tsx#L67 for an example.

I’ll think about what we could do to make this work without this.

sibelius:
my current mock is like this

__mocks__/react-native-safe-area-context.js

```jsx
const inset = {
  top: 0,
  right: 0,
  bottom: 0,
  left: 0,
};

export const SafeAreaConsumer = ({ children }) => {
  return children(inset);
};
```

janicduplessis:
Closing at it seems resolved!

ccfz:
Just to paste the solution from @DnEgorWeb, that he wrote in the another [issue](https://github.com/react-native-community/react-native-safe-area-view/issues/104) here as well:

```
<SafeAreaProvider initialSafeAreaInsets={{ top: 1, left: 2, right: 3, bottom: 4 }}>
  <ComponentUnderTesting />
</SafeAreaProvider>
```

With the above code I did not need any additional mocking.

joaocarmo:
To extend on @sibelius's answer, since I'm also using hooks, I'm mocking this as follows. Hope this helps those searching for solutions to this.

```js
// __mocks__/react-native-safe-area-context.js

const inset = {
  top: 0,
  right: 0,
  bottom: 0,
  left: 0,
}

export const SafeAreaProvider = ({ children }) => children

export const SafeAreaConsumer = ({ children }) => children(inset)

export const useSafeAreaInsets = () => inset
```

gabrielmoncea:
To extend on @joaocarmo answers I'm adding the SafeAreaView component too :)

```javascript
// __mocks__/react-native-safe-area-context.js
import React from 'react'
import { View } from 'react-native';

const inset = {
  top: 0,
  right: 0,
  bottom: 0,
  left: 0,
}

export const SafeAreaProvider = ({ children }) => children

export const SafeAreaConsumer = ({ children }) => children(inset)

export const SafeAreaView = ({ children }) => <View style={inset}>{children}</View>

export const useSafeAreaInsets = () => inset
```

matheusgoc:
[...]

PR #118, opened 2020-07-08
Include tips for testing with enzyme in README

mlodato517:
## Summary

I was hoping to update the README to include the insight from [this comment](https://github.com/th3rdwave/react-native-safe-area-context/issues/31#issuecomment-601074087). My coworkers and I spent some time trying to figure out how to properly render a component under a `SafeAreaProvider` using `enzyme`. One engineer found [this comment](https://github.com/react-navigation/react-native-safe-area-view/issues/104#issuecomment-597789305) which is copied by the above comment. I figure if there are several people independently looking for answers to this across packages it may be worth calling it out in the documentation.

But of course, it may _not_ be worth including documentation for a potential edge case bound to a specific testing utility! And I do realize that [this comment](https://github.com/th3rdwave/react-native-safe-area-context/issues/31#issuecomment-564089209) mentions fixing this more holistically so I'm totally okay with just closing this PR but I didn't know if this would help users discover solutions faster in the interim?

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,6 +263,10 @@ Use `SafeAreaInsetsContext` instead.
 
 Use `initialWindowMetrics` instead.
 
+### initialSafeAreaInsets
+
+Use `initialMetrics` instead.
+
 ## Web SSR
 
 If you are doing server side rendering on the web you can use `initialMetrics` to inject insets and frame value based on the device the user has, or simply pass zero values. Since insets measurement is async it will break rendering your page content otherwise.
--- a/README.md
+++ b/README.md
@@ -263,10 +263,6 @@ Use `SafeAreaInsetsContext` instead.
 
 Use `initialWindowMetrics` instead.
 
-### initialSafeAreaInsets
-
-Use `initialMetrics` instead.
-
 ## Web SSR
 
 If you are doing server side rendering on the web you can use `initialMetrics` to inject insets and frame value based on the device the user has, or simply pass zero values. Since insets measurement is async it will break rendering your page content otherwise.
--- a/README.md
+++ b/README.md
@@ -304,3 +304,17 @@ export function TestSafeAreaProvider() {
   );
 }
 ```
+
+### Testing With Enzyme
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -304,3 +304,17 @@ export function TestSafeAreaProvider() {
...
+`useSafeAreaInsets` or other hooks requiring the `SafeAreaProvider`, you may need to provide
+`initialSafeAreaInsets` to ensure the `SafeAreaProvider` renders its children:
+
+```js
+const wrapper = shallow(
+  <SafeAreaProvider initialSafeAreaInsets={{ top: 0, left: 0, right: 0, bottom: 0 }}
+    <ComponentUnderTest />
+  </SafeAreaProvider>
+)
+```
```
mlodato517, line 320: Not sure if it's worth linking to the appropriate issues/comments in here?

satya164: `initialSafeAreaInsets` is the old API. Aren't you trying to do the same thing as the previous example under testing? What the previous example seems to miss is to render a component inside the provider which you have added here. Can you update the other example to include that instead?

mlodato517: > Can you update the other example to include that

Absolutely! I was wondering what was going on in that example because it didn't make sense to render `SafeAreaProvider` with nothing under it but it seems like it was maybe just an incomplete example?

> `initialSafeAreaInsets` is the old API.

Ah, didn't know that. Should I include it in the "Deprecated apis" section so it's more visible?

satya164: Yeah I think it's just incomplete example. cc @janicduplessis

mlodato517: In the meantime, I've updated the example [here](f79be7369b0a2e0e33981a0eb82d01e803251720) and added `initialSafeAreaInsets` to the deprecated apis list [here](f6d4ee1578c222d30438230e16cd6395d4a84130) 👍

janicduplessis: The deprecated api is actually initialWindowSafeAreaInsets, and is already included in the doc so I just removed it. Rest looks good, thanks!

PR #130, opened 2020-07-26
Fix pointerEvents, other props on SafeAreaView not working on Android

hsource:
## Summary

Because SafeAreaViewManager extends `ViewGroupManager`, some props on `<View>` that are handled by [`ReactViewManager`](https://github.com/facebook/react-native/blob/v0.63.2/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewManager.java) and other classes that extend `ViewGroupManager` aren't handled. Notably, this includes `pointerEvents`, which can be used to make the SafeAreaView ignore taps when it's used as an overlay.

## Test Plan

I installed a [branch](https://github.com/hsource/react-native-safe-area-context/releases/tag/v3.1.1-wanderlog.1) containing the fix to my own app.

To test this, create an instance of `<SafeAreaView style={{ flex: 1 }} pointerEvents="box-none">` that overlays a button or map. The button should be tappable and the map should be pannable.

| Before | After |
| --- | --- |
| Before | After |

### DIFF
--- a/android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaViewManager.java
+++ b/android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaViewManager.java
@@ -6,12 +6,12 @@ import androidx.annotation.Nullable;
 import com.facebook.react.bridge.ReadableArray;
 import com.facebook.react.uimanager.LayoutShadowNode;
 import com.facebook.react.uimanager.ThemedReactContext;
-import com.facebook.react.uimanager.ViewGroupManager;
 import com.facebook.react.uimanager.annotations.ReactProp;
+import com.facebook.react.views.view.ReactViewManager;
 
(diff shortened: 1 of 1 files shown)


janicduplessis:
@Willham12 Should be fixed in 4.1.4

account-mhj:
@janicduplessis SafeAreaProvider has same problems. Properties like `pointerEvent`  wil not works  unless replacing `ViewGroupManager` by `ReactViewManager` in `SafeAreaProviderManager` .

chj-damon:
> @janicduplessis SafeAreaProvider has same problems. Properties like `pointerEvent` wil not work unless replacing `ViewGroupManager` by `ReactViewManager` in `SafeAreaProviderManager` .
@janicduplessis would you mind take a loot at this?

chj-damon:
@account-mhj have you solved this problem?

PR #147, opened 2020-09-17
fix: Xcode 12 compatibility

radko93:
## Summary

Latest Xcode 12 fails to build while without a module to depend on `React-Core` directly hence this change is necessary for all native modules on iOS. This change requires React Native 0.60.2 or newer. For more details please check: https://github.com/facebook/react-native/issues/29633#issuecomment-694187116

## Test Plan

Use this branch to install with an app running on Xcode 12.


janicduplessis (approved):

janicduplessis:
Thanks!!

janicduplessis:
Released in v3.1.8

grit96:
Shouldn't this have been a breaking change as it increases the minimum required `react-native` version?

Issue #126, opened 2020-07-21
Unable to resolve module NativeSafeAreaView


anthowm:
Today I update to last version 3.1.1 and got this issue in both platforms. I tried manual linking and not working. Finally check that in node_modules missing this module
<img width="354" alt="Screenshot 2020-07-21 at 10 29 15" src="https://user-images.githubusercontent.com/5764698/88031775-5c49b780-cb3d-11ea-8554-45316f86a469.png">

mattdamon108:
I have exactly the same issue.

paulg0614:
Same here

woltsu:
`npm start -- --reset-cache` fixed it for me

steveZak:
Still getting this issue, the fix above didn't work for me...

wmonecke:
God this issue made me lose a whole day! Any ideas?

wmonecke:
@janicduplessis

wmonecke:
Current solution:

1) Go to `node_modules/react-native-safe-area-context/src/index.tsx`

2) Add this `export` statement: `export * from './NativeSafeAreaProvider'`

3) `npm start --reset-cache`

4) Keep working.

janicduplessis:
This file doesn't exist anymore (https://github.com/th3rdwave/react-native-safe-area-context/tree/master/src). Would it be possible that you have multiple versions of the library installed? Also you can check what version is actually installed by checking the package.json file "version".

comvenger-brandon:
Having the same problem and I'm sure I'm using the latest react-native-safe-area-context version.

walosha:
I was using expo and this woked for me 

`expo start -c`

florian-milky:
```npm start -- --reset-cache``` was enough for me

FanchenBao:
Using `3.2.0` and `npm start -- --reset-cache` works for me.

ghasemikasra39:
Using `3.3.2` and `npm start -- --reset-cache` also worked for me

karim-belhadjali:
> Current solution:
> 
> 1. Go to `node_modules/react-native-safe-area-context/src/index.tsx`
> 2. Add this `export` statement: `export * from './NativeSafeAreaProvider'`
> 3. `npm start --reset-cache`
> 4. Keep working.

Thank you for this , it did work for me

jacobp100:
This is for v3 of the library, and we're on v4. Please re-open a new issue if you're still facing issues

PR #162, opened 2020-12-09
Fix issue #126

weverkley:


janicduplessis:
Could you explain why this fixes the issue? This module isn't really part of the public API so ideally it shouldn't be exported.

Damoness (commented):
Dont't understand

janicduplessis:
Going to close this, please reopen if you can provide more info on why this change is needed.

PR #166, opened 2021-01-06
set displayName in contexts

dulmandakh:
## Summary

Set displayName for contexts so that they rendered with proper name in React Devtools, thus improve developer experience.

## Test Plan

When you see tree in React Devtools, it'll show SafeAreaInsetsContext.Provider and SafeAreaFrameContext.Provider  instead of just mere Context.Provider.


janicduplessis (approved):
👍 Thanks!

PR #173, opened 2021-01-27
Update Android Tools gradle version

someonewithpc:
## Summary

This was necessary to fix https://github.com/software-mansion/react-native-gesture-handler/issues/494


janicduplessis (approved):
Not sure how this relate to the linked issue, but looks like a good change anyway :)

PR #177, opened 2021-02-06
If view extends beyond window boundaries, do not add that difference to the insets

### LINKED ISSUE #176: Large incorrect insets during rotation on Android
During rotation on Android, `react-native-safe-area-context` can produce very large (varies by device, but in the hundreds) incorrect insets.

This can wind up breaking layouts in various ways, particularly where `ScrollView`s are involved (I'm not _entirely_ certain what makes `ScrollView` so vulnerable).

I have a very simple Snack here: https://snack.expo.io/adtkzgDLf

If you start in portrait, you should see four lines of text. If you rotate to landscape, you should still see four lines of text. After rotating back to portrait, you may only see one line of text (there seem to be timing factors so it may take a couple tries...). It doesn't seem that snack has a view inspector, but when I've used that in other apps I saw that the reason for the missing views was that the views were abnormally tall... I'm guessing because the squashed width from the insets forced the Text elements to wordwrap. (I've seen other UI issues from this as well, the disappearing text elements is merely the simplest to repro/explain)

You can open the Expo Snack log window to see that the right and bottom insets being received are extremely large (they are then corrected back to zero, but it's still enough to create UI problems).

In some cases I've seen the UI recover but there was still a noticeable amount of relayout after rotation that would not normally occur.

This _appears_ due to logic in `SafeAreaUtils.getSafeAreaInsets`:
https://github.com/th3rdwave/react-native-safe-area-context/blob/fc5ec123f05f5da5e02126c9e32057aa05105ead/android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaUtils.java#L59-L60

During the rotation, the `windowWidth`/`windowHeight` are updated to the new orientation before the current view's `height`/`width` are, so the current view's `height`/`width` can wind up being much _larger_ than that of the window. Hence subtracting the window dimensions from the view dimensions still results in a very large number, which then becomes the inset.

I _suspect_ this logic was only intended to handle cases where the current view was much _smaller_ than (or offset from) the window, to avoid applying unnecessary insets? I can't imagine a scenario where you would actually want to _increase_ the insets because the view extended past the window boundary? (even if that was a real issue, correcting it seems outside the responsibilities of this library)

TheAlmightyBob:
Fixes #176

This may not be the ideal fix... I'm open to other options. Perhaps it should consider the scenario of the view extending beyond the window boundaries to simply be "invalid" and return `null`?? Perhaps there's something I'm missing in the intended behavior here?

## Summary

The problem is described in detail in #176. The short version is that, during rotation on Android there's a transitional phase where the window's dimensions have been updated according to the rotation but the view's have not, so the view might extend beyond the height or width of the window. The current calculations in `getSafeAreaInsets` were adding this difference in dimensions to the normal insets and returning extremely large insets.

## Test Plan

I have _not_ tested with a device/app that actually renders under a cutaway or home/back buttons and uses the safe area insets to lay out elements accordingly. This change seems pretty simple/safe in that regard to me, but I'm trusting reviewers/maintainers to say if it looks risky to them. (the status of the "example" app is unclear to me... it doesn't look like it actually uses `react-native-safe-area-context`??)

I _have_ tested with a "normal" device/app, and verified that rotating the device no longer produces super-large insets.

(this code could probably use a comment, but I wasn't sure if I understood it well enough to write a useful one...)

### DIFF
--- a/android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaUtils.java
+++ b/android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaUtils.java
@@ -56,8 +56,8 @@ import androidx.annotation.Nullable;
 
     windowInsets.top = Math.max(windowInsets.top - visibleRect.top, 0);
     windowInsets.left = Math.max(windowInsets.left - visibleRect.left, 0);
-    windowInsets.bottom = Math.max(visibleRect.top + view.getHeight() + windowInsets.bottom - windowHeight, 0);
-    windowInsets.right = Math.max(visibleRect.left + view.getWidth() + windowInsets.right - windowWidth, 0);
+    windowInsets.bottom = Math.max(Math.min(visibleRect.top + view.getHeight() - windowHeight, 0) + windowInsets.bottom, 0);
+    windowInsets.right = Math.max(Math.min(visibleRect.left + view.getWidth() - windowWidth, 0) + windowInsets.right, 0);
     return windowInsets;
   }
 


janicduplessis (approved):
This seems fine, I think the insets should never be bigger than the element it is insetting. This seems consistent with the behavior I've observed on iOS too.

PR #180, opened 2021-02-23
crash on ios10

Liqiankun:


janicduplessis (approved):
Seems like a reasonable change, thanks!!

PR #212, opened 2021-09-01
Fix ReferenceError: React is not defined

rijk:
## Summary

When adding the mock to my Jest setup file following the instructions in the documentation, I get the following error when running my tests:

```error
    ReferenceError: React is not defined

      at Component (node_modules/react-native-safe-area-context/jest/mock.js:23:5)
      at renderWithHooks (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:6016:18)
  
```
## Test Plan

This change makes the tests pass.

### DIFF
--- a/jest/mock.js
+++ b/jest/mock.js
@@ -1,3 +1,5 @@
+import React from 'react'
+
 const MOCK_INITIAL_METRICS = {
   frame: {
     width: 320,


janicduplessis (approved):
Thanks for the fix! I think it was working when I tested it because I’m using the new jsx transform that doesn’t require importing React. This will make it work in both case.

rijk:
Aah, that makes sense. Thanks!

janicduplessis:
published as 3.3.2

PR #227, opened 2021-12-16
New architecture

janicduplessis:
## Test Plan

See updated contributing guide on how to run the example app.

Open the hooks example.


a-eid:
Screen Shot 2021-12-16 at 5 08 20 PM

I'm wondering why we have to use bundler & bundle ?

janicduplessis:
@a-eid I think you need to add `sudo` to the `gem install bundler`. It is to make sure everyone is using the same coocapods version. You can also use your globally installed version if you want by just using `pod install` in example/ios instead.

example/metro.config.js
```diff
@@ -0,0 +1,40 @@
+const path = require('path');
+const exclusionList = require('metro-config/src/defaults/exclusionList');
+const escape = require('escape-string-regexp');
+const pak = require('../package.json');
+
+const root = path.resolve(__dirname, '..');
+
+const modules = Object.keys({
+  ...pak.peerDependencies,
+});
```
jacobp100, line 10:
```suggestion
const modules = Object.keys(pak.peerDependencies);
```

ios/SafeAreaView/RNCSafeAreaContext.mm
```diff
@@ -0,0 +1,71 @@
...
+
+- (NSDictionary *)constantsToExport
+{
+  return [self getConstants];
+}
+
+- (NSDictionary *)getConstants
+{
+  UIWindow *window = [[UIApplication sharedApplication] keyWindow];
+  if (@available(iOS 11.0, *)) {
```
jacobp100, line 28: Is iOS 10 still supported?

ios/SafeAreaView/RNCSafeAreaViewMode.m
```diff
@@ -1,11 +1,15 @@
...
 #import "RNCSafeAreaViewMode.h"
+#import <React/RCTConvert.h>
 
 @implementation RCTConvert (RNCSafeAreaView)
 
-RCT_MULTI_ENUM_CONVERTER(RNCSafeAreaViewMode, (@{
-  @"padding": @(RNCSafeAreaViewModePadding),
-  @"margin": @(RNCSafeAreaViewModeMargin),
-}), RNCSafeAreaViewModePadding, integerValue);
+RCT_MULTI_ENUM_CONVERTER(
```
jacobp100, line 6:
```suggestion
RCT_ENUM_CONVERTER(
```

This was almost definitely my fault 🤣

jacobp100:
This is really cool! I've not really had a chance to play with the new architecture. What's the backwards compat like?

janicduplessis:
@jacobp100 It should be fully backwards compatible. The new architecture is actually already used in RN core.

janicduplessis:
Planning to release this as a prerelease soon to confirm backwards compat is still good. The only thing that might change is require a newer version of react native, but using the new architecture will be opt in still for a while.

ios/SafeAreaView/RNCSafeAreaContext.mm
janicduplessis: It prob was when the code was written, but it shouldn't be now so I can clean that up.

billouboq:
Out of curiosity, what will be the advantages of using the new architecture with this module ? :)

janicduplessis:
For now I'm not using any new feature from the new architecture, but it will enable using rendering priorities which can be nice to have sync rendering when insets change and avoid flickers.

It does enable more code sharing via c++, but the main goal of this PR is just maintain current features.

example/ios/Podfile
```diff
@@ -1,20 +1,33 @@
...
-end
+target 'example' do
+  config = use_native_modules!
+
+  use_react_native!(
+    :path => config[:reactNativePath],
+    :hermes_enabled => true,
+    :fabric_enabled => true,
+    :app_path => "#{__dir__}/..",
+    :config_file_dir => "#{__dir__}/../../..",
```
troZee, line 21:
```suggestion
    :config_file_dir => "#{__dir__}/../..",
```
If `__dir__` equal to PWD command, it will return `system/user/path/react-native-xxxx/example/ios` and then we should go back to times to root folder, which is  `react-native-xxxx` 

Is is correct ?

troZee: Ahh I see your point right now

```
    const codegenConfigFileDir = path.join(confifDir, dependency);
 ```
 So the `fileDir` will be outside the library and then will point to that dependency. But the problem with that solution is that I have the other rn libraries in this dir and the codegen scans those dir, which are outside the project

File: example/package.json [hunk 5]
```diff
@@ -1,22 +1,41 @@
...
-    "react-navigation-tabs": "^2.8.13"
+    "@react-native-community/async-storage": "^1.12.1",
+    "@react-navigation/native": "^6.0.6",
+    "@react-navigation/stack": "^6.0.11",
+    "@react-navigation/bottom-tabs": "^6.0.9",
+    "react": "17.0.2",
+    "react-native": "0.0.0-20211215-2008-f16cbe59e",
+    "react-native-codegen": "^0.0.12",
+    "react-native-gradle-plugin": "^0.0.3",
+    "react-native-safe-area-context": "../"
```
troZee, line 22: I was wondering, why it has been added here, but probably I found a explanation for that 

- the dep is added to package.json
- codegen recognize it as a dep 
- using a custom path the codegen generates needed files 
- in `postInstall` phase this dep is. removed from node modules to avoid duplication 

Could you confirm, if my above thought are correct ?

example/ios/Podfile
janicduplessis: Planning to make some  improvements to config_file_dir. We want to be able to pass a direct path to a library, as well as supporting multiple values with an array.

example/package.json
janicduplessis: Yep that's why, it's a hack to make codegen discover the library properly, while still using the source in the repo with the babel plugin alias to make dev simple.

Not sure if there's a better way to setup that, but for now that works.

example/ios/Podfile
janicduplessis: Interesting, that seems like a better workaround than the one I have now 😅

ios/SafeAreaView/Fabric/RNCSafeAreaViewComponentView.mm
```diff
@@ -0,0 +1,145 @@
...
+  }
+}
+
+- (void)didMoveToWindow
+{
+  _providerView = [self findNearestProvider];
+  [self invalidateSafeAreaInsets];
+}
+
+- (void)invalidateSafeAreaInsets
```
sammy-SC, line 71: Nit: often methods on iOS have suffix "ifNecessary". It suggests to a reader that there is a condition that needs to be met for the action to happen.

Maybe this method could be named "updateStateIfNecessary"?

common/cpp/react/renderer/components/safeareacontext/RNCSafeAreaViewShadowNode.cpp
```diff
@@ -0,0 +1,107 @@
...
+    adjustedStyle.padding()[YGEdgeBottom] = bottom;
+  } else {
+    adjustedStyle.margin()[YGEdgeTop] = top;
+    adjustedStyle.margin()[YGEdgeLeft] = left;
+    adjustedStyle.margin()[YGEdgeRight] = right;
+    adjustedStyle.margin()[YGEdgeBottom] = bottom;
+  }
+
+  auto currentStyle = yogaNode_.getStyle();
+  if (adjustedStyle.padding()[YGEdgeTop] != currentStyle.padding()[YGEdgeTop] ||
```
sammy-SC, line 88: Could we compare `adjustedStyle != currentStyle` instead of comparing every value independently?

Alternatively, compare specifically only padding and margin instead of writing out all edges.

common/cpp/react/renderer/components/safeareacontext/RNCSafeAreaViewShadowNode.cpp
```diff
@@ -0,0 +1,107 @@
...
+  } else {
+    top = valueFromEdges(props.yogaStyle.margin(), YGEdgeTop, YGEdgeVertical);
+    left =
+        valueFromEdges(props.yogaStyle.margin(), YGEdgeLeft, YGEdgeHorizontal);
+    bottom =
+        valueFromEdges(props.yogaStyle.margin(), YGEdgeBottom, YGEdgeVertical);
+    right =
+        valueFromEdges(props.yogaStyle.margin(), YGEdgeRight, YGEdgeHorizontal);
+  }
+  if (std::find(edges.begin(), edges.end(), "top") != edges.end()) {
```
sammy-SC, line 56: I think you could do following:

```
edges.find("top") != edges.end()
```

same result but it is easier to read.

I think free function `std::find` is meant to be used in templates.

sammy-SC (commented):
Look great! I just left few small comments. I was focusing on C++ and iOS part.

ios/SafeAreaView/Fabric/RNCSafeAreaViewComponentView.mm
```diff
@@ -0,0 +1,145 @@
...
+
+  if (UIEdgeInsetsEqualToEdgeInsetsWithThreshold(safeAreaInsets, _currentSafeAreaInsets, 1.0 / RCTScreenScale())) {
+    return;
+  }
+
+  _currentSafeAreaInsets = safeAreaInsets;
+  [self updateState];
+}
+
+- (UIView *)findNearestProvider
```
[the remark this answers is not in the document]
janicduplessis: In the case that there is no provider it will use its own view to get insets from, so I think it is correct to not log anything as it will still work properly.

common/cpp/react/renderer/components/safeareacontext/RNCSafeAreaViewShadowNode.cpp
janicduplessis: I don't think find exists for std::vector, am I missing something?

Ideally this would be a bitmask or set, but that's what is generated by codegen so I can't change it.

common/cpp/react/renderer/components/safeareacontext/RNCSafeAreaViewShadowNode.cpp
janicduplessis: Comaping paddings / margin leads to this compiler error `Invalid operands to binary expression ('IdxRef<YGEdge, &YGStyle::padding_>' and 'IdxRef<YGEdge, &YGStyle::padding_>')`. I assume it doesn't implement the proper compare functions in yoga.

janicduplessis:
@friederbluemle I should have made this change as part of 4.0 in hindsight. Do you still need support for iOS < 11? In that case I would recommend staying on 3.3.x, or you can update the ios version in your podfile.

friederbluemle:
Thank you @janicduplessis :+1:

Issue #92, opened 2020-05-30
Margin mode on iOS can sometimes get stuck with invalid safe area insets


jacobp100:
I had something like the following, 

```js
<SafeAreaView
  style={{ position: 'absolute', left: 0 }}
  edges="left"
  mode="margin"
/>
```

When I rotate to landscape and back to portrait, it gets stuck with the landscape insets

I suspect this is because iOS's safeArea prop does not change if you're outside the safe area insets - which you will be going from landscape to portrait. Maybe we should check if the safeArea changed in `onLayout`

jacobp100:
@janicduplessis I think this could be fixed by emitting an `NSNotification` from `RNCSafeAreaProvider`, and having the views subscribe to that (obviously checking their `_providerView` matches the sender)

Would you accept a PR with this change?

PR #239, opened 2022-02-18
Fix margin mode for absolutely positioned elements

jacobp100:
## Summary

In `SafeAreaViewExample`, add the following change,

```diff
@@ -139,6 +139,17 @@ export default function ReactNativeSafeAreaView() {
           </ScrollView>
         </SafeAreaView>
       </View>
+      <SafeAreaView
+        style={{
+          position: 'absolute',
+          top: 100,
+          left: 0,
+          width: 100,
+          height: 100,
+          backgroundColor: 'red',
+        }}
+        mode="margin"
+      />
     </SafeAreaProvider>
   );
 }
```

Load up an iPhone simulator with a notch, and rotate the device. Observe the view becomes 'stuck'. (See videos)

https://user-images.githubusercontent.com/7275322/154671804-4b004bbb-72f5-4086-88cc-ee763731e038.mov

https://user-images.githubusercontent.com/7275322/154671828-156a24fc-599d-486a-b4f7-dacafa64b85b.mov

This is "expected" - but it'll be much more intuitive if we just use the safe area changed notifications from the provider itself, which should match Android more closely too.

Hopefully it's not too late to smack this into the v4 release - it's technically a breaking change, but if we're fast and call it a bug fix, maybe we can justify it.

## Test Plan

See above

### DIFF
--- a/ios/RNCSafeAreaView.m
+++ b/ios/RNCSafeAreaView.m
@@ -47,16 +47,25 @@ RCT_NOT_IMPLEMENTED(-(instancetype)initWithFrame : (CGRect)frame)
                                     NSStringFromUIEdgeInsets(_currentSafeAreaInsets)];
 }
 
-- (void)safeAreaInsetsDidChange
-{
-  [super safeAreaInsetsDidChange];
-  [self invalidateSafeAreaInsets];
-}
-
 - (void)didMoveToWindow
 {
+  UIView *previousProviderView = _providerView;
   _providerView = [self findNearestProvider];
+  BOOL providerViewDidChange = previousProviderView != _providerView;
+
   [self invalidateSafeAreaInsets];
+
+  if (providerViewDidChange) {
+    [NSNotificationCenter.defaultCenter
+     removeObserver:self
+     name:RNCSafeAreaDidChange
+     object:previousProviderView];
+    [NSNotificationCenter.defaultCenter
+     addObserver:self
+     selector:@selector(safeAreaProviderInsetsDidChange:)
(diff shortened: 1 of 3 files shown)


jacobp100:
Fixes #92

ios/RNCSafeAreaView.m
```diff
@@ -47,16 +47,25 @@ - (NSString *)description
...
 - (void)didMoveToWindow
 {
+  UIView *previousProviderView = _providerView;
   _providerView = [self findNearestProvider];
+  BOOL providerViewDidChange = previousProviderView != _providerView;
+
   [self invalidateSafeAreaInsets];
+
+  if (providerViewDidChange) {
+    [NSNotificationCenter.defaultCenter
```
janicduplessis, line 59: Do we need to cleanup this observer when the view is removed from window too?

janicduplessis (changes requested):
Hey @jacobp100 ! Thanks for the PR

Can you also apply these changes to the fabric implementation (https://github.com/th3rdwave/react-native-safe-area-context/tree/main/ios/Fabric)?

Make sure to rebase on main, I just pushed a fix for the fabric build on iOS. You should be able to test it in the fabric-example app.

ios/RNCSafeAreaView.m
janicduplessis: Ok looks like it is not needed to remove explicitely in dealloc since it keeps weak refs (https://stackoverflow.com/a/36822456). We can keep this as is.

PR #240, opened 2022-02-18
fix: types directory

jeremybarbet:
Types definitions are located in `node_modules/react-native-safe-area-context/lib/typescript`


janicduplessis:
Thanks for the fix!

Issue #251, opened 2022-03-13
Wrong SafeAreaProvider gets imported on react-native-macos


hsjoberg:
Hey.

When importing `<SafeAreaProvider>`, for some reason the native module gets imported instead of the fallback for macOS, `NativeSafeAreaProvider.macos.tsx`.

In other words, `NativeSafeAreaProvider.ts` gets imported instead of `NativeSafeAreaProvider.macos.tsx`, which results in `specs/NativeSafeAreaProvider.js` getting imported causing a crash.
Seems to be some module resolving issue, but it should probably be possible mitigate.

janicduplessis:
Fixed in 4.1.3, thanks for the PR!

PR #252, opened 2022-03-13
Change the name of NativeSafeAreaProvider.ts to NativeSafeAreaProvider.tsx

hsjoberg:
This is because of a module resolving issue causing the wrong file to get imported on react-native-macos.
See the corresponding issue here #251 for more information.


janicduplessis (approved):
Thanks for fixing this!

PR #280, opened 2022-05-11
Update removeEventListner deprecation to just use .remove() 

namrog84:
## Summary

Update deprecrated usage of remove. 

```
The signature '(type: "change", handler: ({ window, screen }: { window: ScaledSize; screen: ScaledSize; }) => void): void' of 'Dimensions.removeEventListener' is deprecated.ts(6387)
index.d.ts(6208, 8): The declaration was marked as deprecated here.
```

## Test Plan

I ran the code, saw the warning. I changed the code. The warning goes away. 
Also we are patch-packaging this change into our own code.

### DIFF
--- a/src/useWindowDimensions.tsx
+++ b/src/useWindowDimensions.tsx
@@ -23,13 +23,13 @@ export default function useWindowDimensions(): DisplayMetrics {
         setDimensions(window);
       }
     }
-    Dimensions.addEventListener('change', handleChange);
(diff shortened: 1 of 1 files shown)


janicduplessis:
Actually I don't think we need this code anymore. It was needed for compatibility with react-native < 0.61, but latest version of this library only supports 0.64+. I will remove the code.

PR #302, opened 2022-08-06
Add required architecture only check

fgagneten:
## Summary
When you are building on android using the new architecture, required architecture check is missing. 

Running the command: `react-native run-android  --active-arch-only` will fail if you don't first build for every single arch (`react-native run-android`)

## Test Plan

1. Try to run `react-native run-android  --active-arch-only` in a project that implements `react-native-safe-area-context` as a first command and it will failed with the following error:

`fcntl(): Bad file descriptor
  /Users/username/Library/Android/sdk/ndk/21.4.7075529/build/core/prebuilt-library.mk:45: *** Android NDK: Aborting    .  Stop.
Android NDK: Check that /Users/username/XX/source/XX-mob/XX/android/app/build/react-ndk/exported/x86/libfb.so exists or that its path is correct --> As I'm running just for `armeabi-v7a` x86 arch will not exist
`

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -82,6 +82,10 @@ android {
                 }
             }
         }
+
+        ndk {
+             abiFilters (*reactNativeArchitectures())
+         }
     }
 
     lintOptions{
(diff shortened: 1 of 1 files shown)


andresribeiro:
I can confirm that this PR fixes #304

janicduplessis (approved):
Thanks! Looks good

PR #305, opened 2022-08-11
React Native 0.69.x Compatibility

brianlenz:
## Summary

Updated the Podspec for iOS so that it supports the latest versions of React Native (`0.69.x`). The dependency needs to be "React-Core" now instead of "React" in order for the iOS build to succeed. Otherwise, there are undefined symbol errors.

## Test Plan

Xcode build failed when packaged with React Native `0.69.4`. With this small change, the Xcode build succeeds and the app runs successfully.


georgeadeishvili (approved):

janicduplessis (approved):
👍 Looks good, thanks!

PR #307, opened 2022-08-23
fix: RCT_NEW_ARCH_ENABLED=0 to disable new arch

nomi9995:
## Summary

RCT_NEW_ARCH_ENABLED=0 is not working, sometimes disturb when we enable and disable the new arch

## Test Plan
`RCT_NEW_ARCH_ENABLED=1` enable new arch and then try to disable `RCT_NEW_ARCH_ENABLED=0` then Xcode will get an error when it will be building app


todorone (approved):

janicduplessis (approved):

janicduplessis:
Looks good, thanks!

PR #312, opened 2022-09-28
feat: support RN 0.70 new arch

janicduplessis:
## Summary

Update to support RN 0.70.0 new arch. This isn't backwards compatible with 0.69.

## Test Plan

Test manually in fabric example app


evelant:
@janicduplessis Seems this is no longer working on 0.70.4 per this issue https://github.com/th3rdwave/react-native-safe-area-context/issues/323, any thoughts?

PR #335, opened 2022-12-10
fix(test): _NativeSafeAreaContext.default.getConstants is not a function

### LINKED ISSUE #255: SafeAreaView with jest breaks
I am getting this error message:
```
TypeError: _NativeSafeAreaContext.default.getConstants is not a function

      at Object.<anonymous> (node_modules/react-native-safe-area-context/lib/commonjs/InitialWindow.native.ts:4:61)
      at Object.<anonymous> (node_modules/react-native-safe-area-context/lib/commonjs/index.tsx:3:1)
```

Running version `4.2.2`

### LINKED ISSUE #290: Breaks using Jest and RN 0.68.2
I upgraded this library today:

```
-    "react-native-safe-area-context": "^3.3.2",
+    "react-native-safe-area-context": "^4.3.1",
```

and found that Jest is unable to run due to an issue with `_NativeSafeAreaContext.default.getConstants`. The error reported is below.

```
 FAIL  __tests__/Initialize.dianastorage.test.tsx
  ● Test suite failed to run

    TypeError: _NativeSafeAreaContext.default.getConstants is not a function

    > 1 | import mockSafeAreaContext from 'react-native-safe-area-context/jest/mock';
        | ^
      2 |
      3 | module.exports = mockSafeAreaContext;
      4 |

      at Object.<anonymous> (node_modules/react-native-safe-area-context/lib/commonjs/InitialWindow.native.js:25:42)
      at Object.<anonymous> (node_modules/react-native-safe-area-context/lib/commonjs/index.tsx:3:1)
      at Object.<anonymous> (node_modules/react-native-safe-area-context/jest/mock.js:20:32)
      at Object.<anonymous> (__mocks__/react-native-safe-area-context.js:1:1)
```

I logged `_NativeSafeAreaContext.default` where the error occurs with `console.log(_NativeSafeAreaContext);`

```tsx
    { default: {} }
```

In Jest, `_NativeSafeAreaContext.default` is `{}` and not `null` or `void` here as the library expects. I don't know whether this is something to do with TurboModules or the new architecture, which I don't have enabled.

andreialecu:
Fixes #290 
Fixes #255 

## Summary

The reason is described in #290, it appears that the native module import comes back as `{ default: {} }`.

The fix ensures `getConstants` is only called if it's defined. When mocked, `default` is `{}` and thus `default.getConstants` is `undefined`.

## Test Plan

Verified to work in our own projects. Using it via `patch-package`

### DIFF
--- a/src/InitialWindow.native.ts
+++ b/src/InitialWindow.native.ts
@@ -1,7 +1,7 @@
 import type { Metrics } from './SafeArea.types';
 import NativeSafeAreaContext from './specs/NativeSafeAreaContext';
 
-export const initialWindowMetrics = (NativeSafeAreaContext?.getConstants()
+export const initialWindowMetrics = (NativeSafeAreaContext?.getConstants?.()
   ?.initialWindowMetrics ?? null) as Metrics | null;
 
 /**


janicduplessis (approved):
Thanks!

PR #340, opened 2023-01-05
Fix `Could not locate shadow view` errors

jacobp100:
> [native] Could not locate shadow view with tag #583, this is probably caused by a temporary inconsistency between native views and shadow views

Caused when a view is removed - presumably the shadow view is removed, and we try and update data

I could reliably reproduce it locally, and can confirm the fix works

I can't quite follow the logic here of why we fallback to using the current view when a provider cannot be found. The current view won't emit `RNCSafeAreaDidChange` events - but then I'm not clear why we even saw this bug in the first place.


jacobp100:
@janicduplessis did you get a chance to look at this?

janicduplessis (approved):

jacobp100:
Thanks!

Issue #316, opened 2022-09-28
Plugin with id 'com.facebook.react' not found.


Asinox:
Any idea how to fix this when `newArchEnabled=true` ?

`Plugin with id 'com.facebook.react' not found.`

Thank you!

janicduplessis:
What version of react native are you using? The lastest version of this library only supports 0.70

efstathiosntonas:
@janicduplessis just to be clear, version `4.4.1` will work only on rn >= 0.70.0?

janicduplessis:
Yes for the new architecture there are too many changes currently to support older versions so for now this library will only support the latest version. I think you can stay on 4.3 for RN 0.69

PR #348, opened 2023-01-19
Update README.md

jacobp100:
https://github.com/th3rdwave/react-native-safe-area-context/issues/316#issuecomment-1262357390


janicduplessis:
I think this chart is for old architecture, which we should still support older RN versions. I added a section that says new arch is only supported for latest RN version but it might not be clear enough as seen by all those issues. Maybe can add a new chart for new arch instead?

janicduplessis (approved):

PR #355, opened 2023-02-19
refactor: reuse 'useSafeAreaInsets' on withSafeAreaInsets

Ne3l:
## Summary

I was profiling my App and found that the stackTrace was quite big

image

Digging into the codebase i found the `withSafeAreaInsets` which renders and extra consumer.

Changed the consumer to the useContext which is the recommended approach according to the react docs and provide 1 less stack and reuse code. 

https://beta.reactjs.org/reference/react/createContext#consumer.

## Test Plan

Tested on my app the change and didn't spot any issue. AFAIK, the useContext is equivalent as the consumer so no more impact.

### DIFF
--- a/src/SafeAreaContext.tsx
+++ b/src/SafeAreaContext.tsx
@@ -116,11 +116,11 @@ const NO_INSETS_ERROR =
   'No safe area value available. Make sure you are rendering `<SafeAreaProvider>` at the top of your app.';
 
 export function useSafeAreaInsets(): EdgeInsets {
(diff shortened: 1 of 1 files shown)


jacobp100:
Thanks for taking the time to look at this

Out of interest, why aren’t you just using the hooks?

This looks safe enough to merge, and we probably should. But we should also consider removing this API

jacobp100 (approved):

janicduplessis (approved):

Issue #204, opened 2021-06-30
Turn off all edges?


duggster:
I would like to be able to pass an empty array to the edges prop of SafeAreaView when I don't want any safe area edges. The use case is a reusable component that on some screens I want it to take into account Safe Area, and other screens already take Safe Area into account and I don't want double padding. So I'd like to pass a prop to my reusable component that can be used like so:
```
const MyComponent = ({ needsSafeArea }) => {
  const edges = [];
  if (needsSafeArea) {
    edges.push('top');
  }

  return (<SafeAreaView edges={edges}>
    {/* My cool component */}    
  </SafeAreaView>);
};
```

This code currently results in all edges being activated when needsSafeArea is false. If there's a better way to handle my use case then please let me know.

fiznool:
If you don't need any edges, can you use a `View` instead?

``` js
const MyComponent = ({ needsSafeArea }) => {
  if (needsSafeArea) {
    return (
      <SafeAreaView edges={['top']}>
        {/* My cool component */}    
      </SafeAreaView>
    );
  }

  return (
    <View>
      {/* My cool component */}    
    </View>
  );
};
```

duggster:
Yes this is a workaround, but not great because you either have to duplicate the code inside or create another layer of abstraction by factoring out the common code into yet another component. It would be much more readable and maintainable to just keep the SafeAreaView structure in place and control the behavior by props. 

An empty edges array seemed intuitive to indicate no edges at first (before I re-read the docs), but probably it would be less disruptive to existing code if an optional flag was added called something like 'disabled'.

CptFabulouso:
I've also come to few use cases where I'd use that passing `edges={[]}` turns off all edges. It's surprising to me it doesn't, because the code clearly says what I want and that is to use no edges. 
If it would work like that and you would have opposite condition like "if something is true, I want only `top` edge, otherwise use default edges"  you could still do that by `edges={condition? ['top'] : undefined}`, so to me it seems like win-win.

Anyway, best workaround I found is to use `useSafeAreaInsets`
```js
import React from 'react';
import { View } from 'react-native';
import { useSafeAreaInsets } from 'react-native-safe-area-context';

const MyComponent = ({ needsSafeArea }) => {
 const { top } = useSafeAreaInsets();
 return (
   <View style={{paddingTop: needsSafeArea? top : 0}}>
     {/* My cool component */}    
   </View>
 );
};
```

MaximeCrp:
[...]

PR #378, opened 2023-04-25
fix: 🐛 empty array sets no edges

VNDRN:
## Summary

As #204 states, providing an empty array as edges to the SafeAreaView component should result into no edges being set. When the edges prop is omitted, all edges should still be set as default

## Test Plan

I've tested this on an iPhone 14 Pro

```Typescript
<SafeAreaView edges={[]} style={{ backgroundColor: 'blue', flex: 1 }}>
      <View style={{ backgroundColor: 'red', flex: 1 }} />
 </SafeAreaView>
```

| no edges (defaults to all) | edges={['bottom']} | edges={[]} |
|--------|--------|--------|
| IMG_3120 | IMG_3121 | IMG_3119 |

### DIFF
--- a/ios/RNCSafeAreaViewEdges.h
+++ b/ios/RNCSafeAreaViewEdges.h
@@ -6,4 +6,5 @@ typedef NS_ENUM(NSInteger, RNCSafeAreaViewEdges) {
   RNCSafeAreaViewEdgesBottom = 0b0010,
   RNCSafeAreaViewEdgesLeft = 0b0001,
   RNCSafeAreaViewEdgesAll = 0b1111,
+  RNCSafeAreaViewEdgesNone = 0b0000,
 };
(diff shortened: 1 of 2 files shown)


VNDRN:
I haven't been able to test if a change needs to be made for android

jacobp100 (approved):
[I think it defaults to all edges on Android](https://github.com/th3rdwave/react-native-safe-area-context/blob/main/android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaView.kt#L33)

It'd be really useful if we could confirm on Android we haven't broken anything. Maybe we could ask in the PR if someone is able to check for us

VNDRN:
I've tried simulating display cutouts with developer options, but couldn't provide any. As this is an iOS only change, it shouldn't affect android though?

jacobp100:
@janicduplessis happy for this to go in?

VNDRN:
I am 😄

janicduplessis:
Looks good, not sure if the same bug exists on android, but if it does we can fix it later, this will only affect iOS.

janicduplessis (approved):

PR #379, opened 2023-04-26
fix:Compatible with version 0.72

NiuGuohui:
## Summary
The Geometry.h file is deprecated and will be removed in the next version of React Native.

### DIFF
--- a/common/cpp/react/renderer/components/safeareacontext/RNCSafeAreaViewState.h
+++ b/common/cpp/react/renderer/components/safeareacontext/RNCSafeAreaViewState.h
@@ -1,7 +1,7 @@
 #pragma once
 
 #include <react/renderer/components/safeareacontext/Props.h>
-#include <react/renderer/graphics/Geometry.h>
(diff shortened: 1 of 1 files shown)


jacobp100:
What are we using from this file?

NiuGuohui:
> What are we using from this file?

I don't know, but it will be an error in RN0.72.

jacobp100:
What versions of RN does this file exist in? Will we be breaking older versions?

janicduplessis:
I think we need to import `RectangleEdges.h` not `Float.h` since we are using the `EdgeInsets` type, this is where it is defined.

I verified and this header was added a very long time ago so it will be backwards compatible.

NiuGuohui:
> I think we need to import `RectangleEdges.h` not `Float.h` since we are using the `EdgeInsets` type, this is where it is defined.
> 
> I verified and this header was added a very long time ago so it will be backwards compatible.

You're right!

janicduplessis (approved):

Issue #295, opened 2022-07-07
Add forwardRef to SafeAreaView.web.tsx


paulpopiel:
https://github.com/th3rdwave/react-native-safe-area-context/blob/ccce005593e19fd24ab27f5e3f9d3f2c00d220c7/src/SafeAreaView.web.tsx#L22

Web version of SafeAreaView is missing a forwardRef passing the ref to the underlying returned View.

e.g. similar to mobile version:
https://github.com/th3rdwave/react-native-safe-area-context/blob/main/src/SafeAreaView.tsx

jacobp100:
Would you be willing to submit a PR?

antoine-cottineau:
Should we close this? I think my PR has implemented what was expected 🙂

PR #383, opened 2023-04-28
Allow web's SafeAreaView to forward ref to its underlying view

antoine-cottineau:
## Summary

This PR solves the issue #295 by adding the ability for the web's `SafeAreaView` to forward the ref to the underlying `View`.

## Test Plan

I run the code by creating a basic Expo app and patching it with my changes.

### DIFF
--- a/src/SafeAreaView.web.tsx
+++ b/src/SafeAreaView.web.tsx
@@ -1,7 +1,11 @@
 import * as React from 'react';
 import { View, StyleSheet } from 'react-native';
 import { useSafeAreaInsets } from './SafeAreaContext';
-import type { Edge, NativeSafeAreaViewProps } from './SafeArea.types';
+import type {
+  Edge,
+  NativeSafeAreaViewInstance,
+  NativeSafeAreaViewProps,
+} from './SafeArea.types';
 
 // prettier-ignore
 const TOP    = 0b1000,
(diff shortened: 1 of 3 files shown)


janicduplessis (changes requested):
Looks good, can you just fix the formatting in SafeAreaView.web.tsx?

antoine-cottineau:
It should be good now 🙂

src/SafeAreaView.web.tsx
```diff
@@ -19,67 +23,66 @@ const edgeBitmaskMap: Record<Edge, number> = {
...
 };
 
-export function SafeAreaView({
-  style = {},
-  mode,
-  edges,
-  ...rest
-}: NativeSafeAreaViewProps) {
-  const insets = useSafeAreaInsets();
+export const SafeAreaView: React.ForwardRefExoticComponent<
```
jacobp100, line 26: Do we need to annotate this type? It should be inferred from forwardRef, and might make the line breaking less aggressive

src/SafeAreaView.web.tsx
```diff
@@ -19,67 +23,66 @@ const edgeBitmaskMap: Record<Edge, number> = {
...
+  NativeSafeAreaViewProps & React.RefAttributes<NativeSafeAreaViewInstance>
+> = React.forwardRef<NativeSafeAreaViewInstance, NativeSafeAreaViewProps>(
+  ({ style = {}, mode, edges, ...rest }, ref) => {
```
jacobp100, line 29: If we take `props` here and destructure it inside the component body, does it avoid the extra indentation?

jacobp100 (approved):
Looks good! Just a few things that might make the formatting less indented

src/SafeAreaView.web.tsx
antoine-cottineau: Yes it removes a level of indentation! I thought we wanted to have the type explicitly written like in `SafeAreaView.tsx`.

src/SafeAreaView.web.tsx
antoine-cottineau: Not needed anymore as removing the return type was enough to remove the extra indentation.

src/SafeAreaView.web.tsx
janicduplessis: Can you also remove it from SafeAreaView.tsx for consistency?

antoine-cottineau: Yes! Done 😉

janicduplessis (approved):
Looks good! Thanks for fixing this

antoine-cottineau:
My pleasure 🙂

janicduplessis:
I might have messed something up, I published 4.5.3 which seems fine. Thanks for finding this.

Issue #308, opened 2022-09-09
"Margin collapse" in SafeAreaView?


fo-fo:
Here's a suggestion. Currently SafeAreaView adds any margin specified by the component user to the safe area margin (assuming mode=margin here.) Sometimes it would be useful to take the _maximum_ of the two margins instead. For example, for a toast at the top of the screen I might want to add some margin myself, but if the safe area margin is already bigger than that margin, then I would like to use that instead.

I know I can already achieve this with `useSafeAreaInsets`, but I'd like to avoid the caveats of the hook.

jacobp100:
I think it's a good feature. Not really sure what the prop would be called - we already used `mode`.

If you're happy to take a stab at it, I'm happy to give pointers.

Start off with this file - https://github.com/th3rdwave/react-native-safe-area-context/blob/main/ios/RNCSafeAreaShadowView.m

jacobp100:
Added in 4.6

PR #394, opened 2023-06-06
Feature: minimum padding

Shaninnik:
## Summary

When building mobile apps I was often faced with the requirement to set a padding on container views (most of the time bottom) that is either equals to bottom safe area or provided value if it is greater than safe area. I've always used something like this:

```
const safeArea = useSafeAreaInsets();
... style={[
          { paddingBottom: Math.max(safeArea.bottom, 24) }
        ]}
```

This works well, but adds big performance impact by using hook. It was extremely noticeable in apps that support device orientation, when changing orientation caused every screen that calls `useSafeAreaInsets` to rerender (sometimes up to 4 times).

This PR adds same functionality on native side. New `minPadding` property on `SafeAreaView` is absolutely optional and PR will not break any existing app. There was a request to add similar feature #308 

I hope someone else find find it useful and it will be merged to main branch.

## Test Plan

New `minPadding` property is added to example app, with a switch to turn it off or on. With it turned on expected outcome is that devices with bottom safe area greater than 24px are not affected, and devices without bottom safe area will have a 24px bottom padding.


Shaninnik:
I might be missing something but I don't see the issue with padding. People who already use padding can continue doing so and ignore this update, they will just get safeArea + padding, the same way as they have it now.

#308 is basically the same: `useful to take the maximum of the two margins instead` but about margin. I chose to use padding because it is default mode. In fact I have never ever used margin mode in any of my apps.

I will re-test with margin mode, it definitely should not be applied with margins.

jacobp100:
> I might be missing something but I don't see the issue with padding. People who already use padding can continue doing so and ignore this update, they will just get safeArea + padding, the same way as they have it now.

I mean if they want the minimum padding behaviour. I’m not sure there’s a use-case for supporting both padding and min-padding at the same time. If we do support both at the same time, the developer is gonna have to think more, and I worry in many cases they’ll get it wrong

I think the real example is floating buttons at the bottom of the display. You need the minimum for the iPhone SE, but don’t want any additional padding on phones without the home buttons. Are there any other cases I missed?

Shaninnik:
Oh I see what you mean now. My reasoning was that padding is something that applies on top of the inset. Initial logic s safeAreaInset + padding, so I just followed it, simply adding additional constraint to safe area insets. I agree that I don't see a use case for minimum padding + padding, but perhaps there is for someone? And it does not overwrite existing logic. I am open to changes though, I can make `minPadding` completely overwrite safeArea+padding if you think this is better.

> I think the real example is floating buttons at the bottom of the display. You need the minimum for the iPhone SE, but don’t want any additional padding on phones without the home buttons. Are there any other cases I missed?

Yeah, this is the exact use case I need it for. And for content inside scroll views.

> Why not? It seems useful

Well.... because it has `padding` name in it...

I think to avoid both issues you mentioned perhaps I should leave the logic exactly the same as it is and just rename `minPadding` to something better? For example `minSafeArea`. Then the logic is:

```
Padding mode:
SafeAreaView padding = max(minSafeArea, safeArea) + padding

Margin mode:
SafeAreaView margin = max(minSafeArea, safeArea) + margin
```

I should also update docs to mention new property.

What do you think?

jacobp100:
Sorry - I just realised that linked issue didn't actually explicitly demonstrate an API. One example would be to make the `edges` prop (optionally) take an object:-

```js
<SafeAreaView
  style={{ padding: ... }}
  // Naming TBC - just for example
  edges={{ top: 'off', left: 'additive', right: 'additive', bottom: 'maximum' }}
/>
```

When you pass in an array (the current API), every value in the array would be `additive`, and every other value would be `off`

Let me know what you think! I'm happy to look at other APIs too. It's worth spending the time now, because once this gets merged we're gonna be stuck with that decision 🤣

Shaninnik:
I wanted solution to be as less intrusive for others as possible, and my proposed one is, most people can just ignore it and nothing will change for them. For people who are trying to achieve the same thing with `useSafeAreaInsets` and `{ paddingBottom: Math.max(safeArea.bottom, 24) }` this will be very a very familiar logic and I am sure they will be comfortable with it, they also are not forced to upgrade, they can continue doing it this way as it works and just causes some slow rerenders on device orientation changes. People who currently use padding with SafeAreaView and live with increased paddings on devices with safe area most probably not going to change it because their design is already based on the fact of double padding and it does not bother them (I think this is the case you are worried about)

Changing `edges` property from array to object is not an option, it is a breaking change and I am 100% sure that we don't want that. Having it as a union `Edges[] | {top: 'off' | 'additive' | 'maximum', left:..., right:..., bottom:...}` I think is the same level of confusing as dealing with padding and you will have to list all edges all the time. For example I have a use use case right now where I need only a bottom edge from safe area with minimum padding. Instead of being like this:

```
// my proposal. 100% agree that `minPadding` from my initial PR is a bad name, hence `minSafeArea`
<SafeAreaView edges={['bottom']} minSafeArea={{ bottom: 24 }}> 
```

I will have to be like this:
```
// your proposal
<SafeAreaView style={{paddingBottom: 24}} edges={{ top: 'off', left: 'off', right: 'off', bottom: 'maximum' }}> 
```

Honestly I would rather stick with existing implementation of `edges` as I (and everyone) got already used to it and it does not force you to list every edge. Also dealing with native bridge for union already gives me nightmares.

Other options, all naming is just an example 
```
/// edgeLogic: {{top?: 'additive' | 'maximum'...}} - if no value supplied default is 'additive'
<SafeAreaView style={{paddingBottom: 24}} edges={['bottom']} edgeLogic={{bottom: 'maximum'}}>  
```

```
/// edgeLogic: 'maximum' | 'additive' - default 'additive'. Very simple option but will apply same logic for all edges
<SafeAreaView style={{paddingBottom: 24}} edges={['bottom']} edgeLogic={'maximum}> 
```

This is a fairly simple PR for a fairly simple and very niche feature. There was only one other person who complained, and this lib is used by millions. I don't think we should overcomplicate it. I am ready to make whatever change you think is the best and get it merged sooner rather than spending weeks talking about it 😅

Shaninnik:
It 100% does and this is exactly how imagined it. The downside is that default has to be 'additive', because it is what will be used if you pass array, and if I want to disable all edges except one I will have to explicitly set 'off' for everything:

```
edges={{ top: 'off', left: 'off', right: 'off', bottom: 'maximum' }}
```

This is the exact opposite of how current array edges property works.

Shaninnik:
@jacobp100 I have started doing changes we discussed, it is a bit more involved than my initial PR. I've made a work in progress commit with JS and Android parts almost done. Just need to sort out iOS part, I admit I am not very good with Objective C so need to spend some time.

A couple of notes:

1) I don't think we need 'off' in JS part in `type EdgeMode = 'off' | 'additive' | 'maximum';` since we agreed that they default to 'off' if not passed in object. It is enough to have  `type EdgeMode = 'additive' | 'maximum';` and passing something like `edges={{bottom: 'maximum'}}`

2) I don't think we need `useMemo` in SafeAreaView. It is not expensive calc and useMemo will depend on `edges` property anyway and I don't expect people are memoing props they pass to SafeAreaView

```const nativeEdges = Array.isArray(edges)
    ? edges.reduce<EdgeRecord>((accum, edge) => {
        accum[edge] = 'additive';
        return accum;
      }, {})
    : edges;```

3) I am not sure about web version of SafeAreaView, does it even need to support 'maximum' mode?

File: src/SafeAreaView.tsx [hunk 1]
```diff
@@ -11,12 +12,24 @@ export const SafeAreaView = React.forwardRef<
   NativeSafeAreaViewInstance,
   SafeAreaViewProps
 >(({ edges, ...props }, ref) => {
+  const nativeEdges = Array.isArray(edges)
+    ? edges.reduce<EdgeRecord>((accum, edge) => {
+        accum[edge] = 'additive';
+        return accum;
+      }, {})
+    : edges;
```
jacobp100, line 20:
```suggestion
  const nativeEdges = useMemo(() => {
    return Array.isArray(edges)
      ? edges.reduce<EdgeRecord>((accum, edge) => {
          accum[edge] = 'additive';
          return accum;
        }, {})
      : (edges ?? {
           top: 'additive',
           left: 'additive',
           bottom: 'additive',
           right: 'additive',
         });
    })
```

android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaViewShadowNode.kt
```diff
@@ -65,20 +66,20 @@ class SafeAreaViewShadowNode : LayoutShadowNode() {
...
+    val insetTop: Float = if (edges.top != SafeAreaViewEdgeModes.OFF ) insets.top else 0.0f
+    val insetRight: Float = if (edges.right != SafeAreaViewEdgeModes.OFF) insets.right else 0.0f
+    val insetBottom: Float = if (edges.bottom != SafeAreaViewEdgeModes.OFF) insets.bottom else 0.0f
+    val insetLeft: Float = if (edges.left != SafeAreaViewEdgeModes.OFF) insets.left else 0.0f
     if (localData.mode == SafeAreaViewMode.PADDING) {
-      super.setPadding(Spacing.TOP, insetTop + top)
-      super.setPadding(Spacing.RIGHT, insetRight + right)
-      super.setPadding(Spacing.BOTTOM, insetBottom + bottom)
-      super.setPadding(Spacing.LEFT, insetLeft + left)
+      super.setPadding(Spacing.TOP,  if (edges.top == SafeAreaViewEdgeModes.MAXIMUM ) max(insetTop, top) else insetTop + top)
```
jacobp100, line 74: It might be easier to make a function like `applyEdges(edgeMode: SafeAreaViewEdgeModes, edgeValue: float, insetValue: float)` - then we can get rid of a fair few lines here

jacobp100:
For 1 - the user needs to be able to do it conditionally - e.g. `{ top: condition ? 'off' : 'maximum' }`

I'm happy to help with iOS stuff. Let me know if you have any questions

Shaninnik:
For 1: is anything wrong with `{top: condition ? undefined : 'maximum'}`? or `condition ? {} : { top : 'maximum'}` Undefined is already an off state, same as not passing an edge in the array

For 2: Maybe, but it will not work on JS side. I expect 99.9% of users will use it like that (passing edges property inline):

```
<SafeAreaView
      edges={{ bottom: 'maximum' }}
    >
        
```

This nets a new object reference for each rerender and useMemo will be useless, it will not memo anything and will run every render creating new reference for `nativeEdges`:

Screen Shot 2023-06-08 at 8 42 30 AM

I am unsure about what degree of memoization RN has for passing props to native components, if it is a deep compare then it literally does not matter if we useMemo or not, if it by reference and it prevents native component updates then we need edgeBitmask trick similar to SafeAreaView.web or React.memo the entire SafeAreaView. But I think it is out of scope of this PR as it will be additional memoization that is not present in current version of the library. I can do it as separate PR after we merge this one.

This does work if users will be memoing edges property:
```
const edges = React.useMemo(() => {
    return { bottom: 'additive' as const };
  }, []);
  
<SafeAreaView
      edges={edges}
    >
```
But I don't think anyone is really doing it that way. Am I wrong?

For 3: I would rather skip it for now as I have never ever even tried react native web before. I can investigate it after we merge and create another PR if it needs some additions to support `maximum` mode.

> I'm happy to help with iOS stuff. Let me know if you have any questions

Going to try it today. I could easily do it in Swift but Objective C... I am sure there will be a lot of swearing involved

jacobp100:
1 - it matches stuff like `contentInset` on ScrollViews, missing edges are set to zero, but you can still set defined edges to zero. Not sure about Android, but `RCT_ENUM_CONVERTER` makes it really easy to default to `off` when it's not defined

2 - I think people would define it outside the component, rather than `useMemo`

jacobp100:
```js
const edges = ['top']

const Component = () => {
  return <SafeAreaView edges={edges} />
}
```

Shaninnik:
I am probably missing what you are trying to achieve:

Screen Shot 2023-06-08 at 10 08 35 AM

jacobp100:
Yeah - you gotta move that `_edges` variable declaration completely outside the component - before `export default function ReactNativeSafeAreaView()`

Shaninnik:
Yeah, right, if `_edges` is outside then it has a stable reference and `useMemo` will work in that case. My point is, are library users doing it that way? I don't think so. I always use it inline:

```
<SafeAreaView edges={['right', 'bottom', 'left']} />
```

Official docs form SafeAreaView use it inline. I looked at the `react-navigation` source code and they use it inline too. And I expect almost everyone uses it inline. In that case, like I have said in my comments, `useMemo` is very misleading and not doing what we are trying to achieve.

I am happy to improve SafeAreaView memoization in a separate PR.

jacobp100:
A quick search - yes people do it

https://github.com/mattermost/mattermost-mobile/blob/8d475d8a36dae72334a5950d30db217deea72fc2/app/screens/code/index.tsx#L7

https://github.com/tsyirvo/react-native-starter/blob/695628c36dc8391444c77bc5b3f53646953a3031/src/components/ui/SafeView.tsx#L2

https://github.com/AaronZhangL/mobile-native/blob/63232c195736a825a7f52e514eec445d6219790d/src/discovery/v2/PlusDiscoveryScreen.tsx#L40

Shaninnik:
I think it worth mentioning it in docs in Optimization section then?

Or better, properly memoize SafeAreaView with `React.memo`?

jacobp100:
RN already does some memoization - if the prop identity stays the same, the native code handler for the prop isn't called. We don't need to go any further than adding a `useMemo` for the people who do use stable identities

android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaViewShadowNode.kt
```diff
@@ -65,23 +66,27 @@ class SafeAreaViewShadowNode : LayoutShadowNode() {
...
-      super.setMargin(Spacing.LEFT, insetLeft + left)
+      super.setMargin(Spacing.TOP, getEdgeValue(edges.top, insetTop, top))
+      super.setMargin(Spacing.RIGHT, getEdgeValue(edges.right, insetRight, right))
+      super.setMargin(Spacing.BOTTOM, getEdgeValue(edges.bottom, insetBottom, bottom))
+      super.setMargin(Spacing.LEFT, getEdgeValue(edges.left, insetLeft, left))
     }
   }
 
+  private fun getEdgeValue(edgeMode: SafeAreaViewEdgeModes, insetValue: Float, edgeValue: Float): Float {
+    return if (edgeMode == SafeAreaViewEdgeModes.MAXIMUM ) max(insetValue, edgeValue) else insetValue + edgeValue
```
jacobp100, line 87: You could check for `edgeMode === SafeAreaViewEdgeModes.OFF` here too and remove lines 69-72

ios/RNCSafeAreaShadowView.m
```diff
@@ -108,27 +109,41 @@ - (void)updateInsets
...
   }
 }
 
+- (CGFloat)getEdgeValue:(RNCSafeAreaViewEdgeModes) edgeMode insetValue:(CGFloat)insetValue edgeValue:(CGFloat) edgeValue
+{
+    if (edgeMode == RNCSafeAreaViewEdgeModesMaximum) {
+        return MAX(insetValue, edgeValue);
+    } else {
+        return insetValue + edgeValue;
+    }
```
jacobp100, line 140: In the sidebar to the right, you can set the tab size to 2

```suggestion
  if (edgeMode == RNCSafeAreaViewEdgeModesMaximum) {
    return MAX(insetValue, edgeValue);
  } else {
    return insetValue + edgeValue;
  }
```

ios/RNCSafeAreaViewEdgeModes.h
```diff
@@ -0,0 +1,11 @@
+#import <Foundation/Foundation.h>
+
+typedef NS_ENUM(NSInteger, RNCSafeAreaViewEdgeModes) {
+    RNCSafeAreaViewEdgeModesAdditive = 0b1000,
```
jacobp100, line 4: It might be better to leave these as the defaults since you can't use bitwise operations on this type. Also, if you put `off` first, it'll have a value 0

ios/RNCSafeAreaViewEdgeModes.m
```diff
@@ -0,0 +1,29 @@
+#import "RNCSafeAreaViewEdgeModes.h"
+#import <React/RCTConvert.h>
+
+@implementation RCTConvert (RNCSafeAreaView)
+
+RCT_MULTI_ENUM_CONVERTER(
```
jacobp100, line 6:
```suggestion
RCT_ENUM_CONVERTER(
```

ios/RNCSafeAreaViewEdgeModes.m
```diff
@@ -0,0 +1,29 @@
...
+
+@implementation NSString (EnumParser)
+
+- (RNCSafeAreaViewEdgeModes)RNCSafeAreaViewEdgeModesFromString{
+    NSDictionary<NSString*,NSNumber*> *edgeModes = @{
+                            @"off": @(RNCSafeAreaViewEdgeModesOff),
+                            @"additive": @(RNCSafeAreaViewEdgeModesAdditive),
+                            @"maximum": @(RNCSafeAreaViewEdgeModesMaximum),
+                            };
+    return edgeModes[self].integerValue;
```
jacobp100, line 26: This is what RCT_ENUM_CONVERTER does - with additional logic to avoid recreating the dict etc.

ios/RNCSafeAreaViewEdgeModes.h
```diff
@@ -0,0 +1,11 @@
+#import <Foundation/Foundation.h>
+
+typedef NS_ENUM(NSInteger, RNCSafeAreaViewEdgeModes) {
```
jacobp100, line 3:
```suggestion
typedef NS_ENUM(NSInteger, RNCSafeAreaViewEdgeMode) {
```

src/SafeAreaView.tsx
```diff
@@ -1,22 +1,38 @@
...
+  }, [edges]);
+
   return (
     <NativeSafeAreaView
       {...props}
-      // Codegen doesn't support default values for array types so
-      // set it here.
-      edges={edges ?? ['bottom', 'left', 'right', 'top']}
+      edges={
+        nativeEdges ?? {
```
jacobp100, line 29: Can we move this object into the useMemo - or even better, outside the component - so we keep stable references? I know it was that way before, but we have the chance to fix it

ios/RNCSafeAreaViewEdgeModes.m
jacobp100: The multi enum converter is for bitmasks

jacobp100:
@Shaninnik Looking good! Added a few comments. Thanks for working on this too - it'll really help people out

src/specs/NativeSafeAreaView.ts
```diff
@@ -1,11 +1,11 @@
 import codegenNativeComponent from 'react-native/Libraries/Utilities/codegenNativeComponent';
 import type { WithDefault } from 'react-native/Libraries/Types/CodegenTypes';
 import type { ViewProps, HostComponent } from 'react-native';
+import { EdgeRecord } from '../SafeArea.types';
 
 export interface NativeProps extends ViewProps {
   mode?: WithDefault<'padding' | 'margin', 'padding'>;
-  // Should be `'top' | 'right' | 'bottom' | 'left'` but that generates invalid code.
-  edges?: readonly string[];
+  edges?: EdgeRecord;
```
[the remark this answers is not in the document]
Shaninnik: Yeah you right. I though codegen worked like that but it does not. I've tried with:

```
edges?: Readonly<{
    top?: WithDefault<'off' | 'additive' | 'maximum', 'off'>;
    right?: WithDefault<'off' | 'additive' | 'maximum', 'off'>;
    bottom?: WithDefault<'off' | 'additive' | 'maximum', 'off'>;
    left?: WithDefault<'off' | 'additive' | 'maximum', 'off'>;
  }>;
  ``` 
  
 It seems like codegen follows this but now I get a series of errors in cpp files:
 
 ```
 /Users/nikitashanin/projects/react-native-safe-area-context/common/cpp/react/renderer/components/safeareacontext/RNCSafeAreaViewShadowNode.cpp:60:38: error: no member named 'end' in 'facebook::react::RNCSafeAreaViewEdgesStruct'
  if (std::find(edges.begin(), edges.end(), "left") != edges.end()) {
```
This is the first time I am trying new architecture and documentation on it is still not very clear. It looks like it still tries to work with `edges` property as if it was an array, but I think I am not supposed to make any changes in these cpp files myself, am I?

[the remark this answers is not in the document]
jacobp100: I think the old type was `string` because otherwise it failed. You might need `{ top: string, left: string ... }`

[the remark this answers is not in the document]
Shaninnik: It dies not work this way `{ top: string, left: string ... }` either, same error.

```
5: Task failed with an exception.
-----------
* What went wrong:
Execution failed for task ':app:buildCMakeDebug[arm64-v8a]'.
> com.android.ide.common.process.ProcessException: ninja: Entering directory `/Users/nikitashanin/projects/react-native-safe-area-context/example/android/app/.cxx/Debug/69p2a451/arm64-v8a'
  [0/2] Re-checking globbed directories...
  [1/2] Re-running CMake...
  -- Configuring done
  -- Generating done
  -- Build files have been written to: /Users/nikitashanin/projects/react-native-safe-area-context/example/android/app/.cxx/Debug/69p2a451/arm64-v8a
  [0/2] Re-checking globbed directories...
  [1/5] Building CXX object safeareacontext_autolinked_build/CMakeFiles/react_codegen_safeareacontext.dir/Users/nikitashanin/projects/react-native-safe-area-context/example/node_modules/react-native-safe-area-context/common/cpp/react/renderer/components/safeareacontext/RNCSafeAreaViewShadowNode.cpp.o
  FAILED: safeareacontext_autolinked_build/CMakeFiles/react_codegen_safeareacontext.dir/Users/nikitashanin/projects/react-native-safe-area-context/example/node_modules/react-native-safe-area-context/common/cpp/react/renderer/components/safeareacontext/RNCSafeAreaViewShadowNode.cpp.o 
  /Users/nikitashanin/Library/Android/sdk/ndk/23.1.7779620/toolchains/llvm/prebuilt/darwin-x86_64/bin/clang++ --target=aarch64-none-linux-android21 --gcc-toolchain=/Users/nikitashanin/Library/Android/sdk/ndk/23.1.7779620/toolchains/llvm/prebuilt/darwin-x86_64 --sysroot=/Users/nikitashanin/Library/Android/sdk/ndk/23.1.7779620/toolchains/llvm/prebuilt/darwin-x86_64/sysroot -Dreact_codegen_safeareacontext_EXPORTS -I/Users/nikitashanin/projects/react-native-safe-area-context/example/node_modules/react-native-safe-area-context/android/src/main/jni/. -I/Users/nikitashanin/projects/react-native-safe-area-context/example/node_modules/react-native-safe-area-context/android/src/main/jni/../../../../common/cpp -I/Users/nikitashanin/projects/react-native-safe-area-context/example/node_modules/react-native-safe-area-context/android/src/main/jni/../../../build/generated/source/codegen/jni -I/Users/nikitashanin/projects/react-native-safe-area-context/example/node_modules/react-native-safe-area-context/android/src/main/jni/../../../build/generated/source/codegen/jni/react/renderer/components/safeareacontext -isystem /Users/nikitashanin/.gradle/caches/transforms-3/4223004ae9355336a9d95bfd0b9550ed/transformed/jetified-fbjni-0.3.0/prefab/modules/fbjni/include -isystem /Users/nikitashanin/.gradle/caches/transforms-3/84bec54958abf3a45efcfc6e26585866/transformed/jetified-react-android-0.71.7-debug/prefab/modules/folly_runtime/include -isystem /Users/nikitashanin/.gradle/caches/transforms-3/84bec54958abf3a45efcfc6e26585866/transformed/jetified-react-android-0.71.7-debug/prefab/modules/glog/include -isystem /Users/nikitashanin/.gradle/caches/transforms-3/84bec54958abf3a45efcfc6e26585866/transformed/jetified-react-android-0.71.7-debug/prefab/modules/jsi/include -isystem /Users/nikitashanin/.gradle/caches/transforms-3/84bec54958abf3a45efcfc6e26585866/transformed/jetified-react-android-0.71.7-debug/prefab/modules/react_codegen_rncore/include -isystem /Users/nikitashanin/.gradle/caches/transforms-3/84bec54958abf3a45efcfc6e26585866/transformed/jetified-react-android-0.71.7-debug/prefab/modules/react_debug/include -isystem /Users/nikitashanin/.gradle/caches/transforms-3/84bec54958abf3a45efcfc6e26585866/transformed/jetified-react-android-0.71.7-debug/prefab/modules/react_nativemodule_core/include -isystem /Users/nikitashanin/.gradle/caches/transforms-3/84bec54958abf3a45efcfc6e26585866/transformed/jetified-react-android-0.71.7-debug/prefab/modules/react_render_core/include -isystem /Users/nikitashanin/.gradle/caches/transforms-3/84bec54958abf3a45efcfc6e26585866/transformed/jetified-react-android-0.71.7-debug/prefab/modules/react_render_debug/include -isystem /Users/nikitashanin/.gradle/caches/transforms-3/84bec54958abf3a45efcfc6e26585866/transformed/jetified-react-android-0.71.7-debug/prefab/modules/react_render_graphics/include -isystem /Users/nikitashanin/.gradle/caches/transforms-3/84bec54958abf3a45efcfc6e26585866/transformed/jetified-react-android-0.71.7-debug/prefab/modules/react_render_mapbuffer/include -isystem /Users/nikitashanin/.gradle/caches/transforms-3/84bec54958abf3a45efcfc6e26585866/transformed/jetified-react-android-0.71.7-debug/prefab/modules/rrc_view/include -isystem /Users/nikitashanin/.gradle/caches/transforms-3/84bec54958abf3a45efcfc6e26585866/transformed/jetified-react-android-0.71.7-debug/prefab/modules/turbomodulejsijni/include -isystem /Users/nikitashanin/.gradle/caches/transforms-3/84bec54958abf3a45efcfc6e26585866/transformed/jetified-react-android-0.71.7-debug/prefab/modules/yoga/include -g -DANDROID -fdata-sections -ffunction-sections -funwind-tables -fstack-protector-strong -no-canonical-prefixes -D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security   -fno-limit-debug-info  -fPIC -fexceptions -frtti -std=c++17 -Wall -Wpedantic -Wno-gnu-zero-variadic-macro-arguments -DLOG_TAG=\"ReactNative\" -DFOLLY_NO_CONFIG=1 -DFOLLY_HAVE_CLOCK_GETTIME=1 -DFOLLY_USE_LIBCPP=1 -DFOLLY_MOBILE=1 -DFOLLY_HAVE_RECVMMSG=1 -DFOLLY_HAVE_PTHREAD=1 -DFOLLY_HAVE_XSI_STRERROR_R=1 -MD -MT safeareacontext_autolinked_build/CMakeFiles/react_codegen_safeareacontext.dir/Users/nikitashanin/projects/react-native-safe-area-context/example/node_modules/react-native-safe-area-context/common/cpp/react/renderer/components/safeareacontext/RNCSafeAreaViewShadowNode.cpp.o -MF safeareacontext_autolinked_build/CMakeFiles/react_codegen_safeareacontext.dir/Users/nikitashanin/projects/react-native-safe-area-context/example/node_modules/react-native-safe-area-context/common/cpp/react/renderer/components/safeareacontext/RNCSafeAreaViewShadowNode.cpp.o.d -o safeareacontext_autolinked_build/CMakeFiles/react_codegen_safeareacontext.dir/Users/nikitashanin/projects/react-native-safe-area-context/example/node_modules/react-native-safe-area-context/common/cpp/react/renderer/components/safeareacontext/RNCSafeAreaViewShadowNode.cpp.o -c /Users/nikitashanin/projects/react-native-safe-area-context/example/node_modules/react-native-safe-area-context/common/cpp/react/renderer/components/safeareacontext/RNCSafeAreaViewShadowNode.cpp
  In file included from /Users/nikitashanin/projects/react-native-safe-area-context/example/node_modules/react-native-safe-area-context/common/cpp/react/renderer/components/safeareacontext/RNCSafeAreaViewShadowNode.cpp:1:
  In file included from /Users/nikitashanin/projects/react-native-safe-area-context/example/node_modules/react-native-safe-area-context/common/cpp/react/renderer/components/safeareacontext/RNCSafeAreaViewShadowNode.h:6:
  /Users/nikitashanin/projects/react-native-safe-area-context/example/node_modules/react-native-safe-area-context/android/src/main/jni/../../../../common/cpp/react/renderer/components/safeareacontext/RNCSafeAreaViewState.h:19:7: warning: designated initializers are a C++20 extension [-Wc++20-designator]
        .left = (float)value["left"].getDouble(),
        ^
  /Users/nikitashanin/projects/react-native-safe-area-context/example/node_modules/react-native-safe-area-context/common/cpp/react/renderer/components/safeareacontext/RNCSafeAreaViewShadowNode.cpp:56:23: error: no member named 'begin' in 'facebook::react::RNCSafeAreaViewEdgesStruct'
    if (std::find(edges.begin(), edges.end(), "top") != edges.end()) {
                  ~~~~~ ^
  /Users/nikitashanin/projects/react-native-safe-area-context/example/node_modules/react-native-safe-area-context/common/cpp/react/renderer/components/safeareacontext/RNCSafeAreaViewShadowNode.cpp:56:38: error: no member named 'end' in 'facebook::react::RNCSafeAreaViewEdgesStruct'
    if (std::find(edges.begin(), edges.end(), "top") != edges.end()) {
                                 ~~~~~ ^
  /Users/nikitashanin/projects/react-native-safe-area-context/example/node_modules/react-native-safe-area-context/common/cpp/react/renderer/components/safeareacontext/RNCSafeAreaViewShadowNode.cpp:56:61: error: no member named 'end' in 'facebook::react::RNCSafeAreaViewEdgesStruct'
    if (std::find(edges.begin(), edges.end(), "top") != edges.end()) {
                                                        ~~~~~ ^
  /Users/nikitashanin/projects/react-native-safe-area-context/example/node_modules/react-native-safe-area-context/common/cpp/react/renderer/components/safeareacontext/RNCSafeAreaViewShadowNode.cpp:60:23: error: no member named 'begin' in 'facebook::react::RNCSafeAreaViewEdgesStruct'
    if (std::find(edges.begin(), edges.end(), "left") != edges.end()) {
                  ~~~~~ ^
  /Users/nikitashanin/projects/react-native-safe-area-context/example/node_modules/react-native-safe-area-context/common/cpp/react/renderer/components/safeareacontext/RNCSafeAreaViewShadowNode.cpp:60:38: error: no member named 'end' in 'facebook::react::RNCSafeAreaViewEdgesStruct'
    if (std::find(edges.begin(), edges.end(), "left") != edges.end()) {
                                 ~~~~~ ^
  /Users/nikitashanin/projects/react-native-safe-area-context/example/node_modules/react-native-safe-area-context/common/cpp/react/renderer/components/safeareacontext/RNCSafeAreaViewShadowNode.cpp:60:62: error: no member named 'end' in 'facebook::react::RNCSafeAreaViewEdgesStruct'
    if (std::find(edges.begin(), edges.end(), "left") != edges.end()) {
                                                         ~~~~~ ^
  /Users/nikitashanin/projects/react-native-safe-area-context/example/node_modules/react-native-safe-area-context/common/cpp/react/renderer/components/safeareacontext/RNCSafeAreaViewShadowNode.cpp:64:23: error: no member named 'begin' in 'facebook::react::RNCSafeAreaViewEdgesStruct'
    if (std::find(edges.begin(), edges.end(), "right") != edges.end()) {
                  ~~~~~ ^
  /Users/nikitashanin/projects/react-native-safe-area-context/example/node_modules/react-native-safe-area-context/common/cpp/react/renderer/components/safeareacontext/RNCSafeAreaViewShadowNode.cpp:64:38: error: no member named 'end' in 'facebook::react::RNCSafeAreaViewEdgesStruct'
    if (std::find(edges.begin(), edges.end(), "right") != edges.end()) {
                                 ~~~~~ ^
  /Users/nikitashanin/projects/react-native-safe-area-context/example/node_modules/react-native-safe-area-context/common/cpp/react/renderer/components/safeareacontext/RNCSafeAreaViewShadowNode.cpp:64:63: error: no member named 'end' in 'facebook::react::RNCSafeAreaViewEdgesStruct'
    if (std::find(edges.begin(), edges.end(), "right") != edges.end()) {
                                                          ~~~~~ ^
  /Users/nikitashanin/projects/react-native-safe-area-context/example/node_modules/react-native-safe-area-context/common/cpp/react/renderer/components/safeareacontext/RNCSafeAreaViewShadowNode.cpp:68:23: error: no member named 'begin' in 'facebook::react::RNCSafeAreaViewEdgesStruct'
    if (std::find(edges.begin(), edges.end(), "bottom") != edges.end()) {
                  ~~~~~ ^
  /Users/nikitashanin/projects/react-native-safe-area-context/example/node_modules/react-native-safe-area-context/common/cpp/react/renderer/components/safeareacontext/RNCSafeAreaViewShadowNode.cpp:68:38: error: no member named 'end' in 'facebook::react::RNCSafeAreaViewEdgesStruct'
    if (std::find(edges.begin(), edges.end(), "bottom") != edges.end()) {
                                 ~~~~~ ^
  /Users/nikitashanin/projects/react-native-safe-area-context/example/node_modules/react-native-safe-area-context/common/cpp/react/renderer/components/safeareacontext/RNCSafeAreaViewShadowNode.cpp:68:64: error: no member named 'end' in 'facebook::react::RNCSafeAreaViewEdgesStruct'
    if (std::find(edges.begin(), edges.end(), "bottom") != edges.end()) {
                                                           ~~~~~ ^
  1 warning and 12 errors generated.
  [2/5] Building CXX object CMakeFiles/appmodules.dir/OnLoad.cpp.o
  [3/5] Building CXX object CMakeFiles/appmodules.dir/Users/nikitashanin/projects/react-native-safe-area-context/example/android/app/build/generated/rncli/src/main/jni/rncli.cpp.o
  ninja: build stopped: subcommand failed.
  
  C++ build system [build] failed while executing:
      /Users/nikitashanin/Library/Android/sdk/cmake/3.18.1/bin/ninja \
        -C \
        /Users/nikitashanin/projects/react-native-safe-area-context/example/android/app/.cxx/Debug/69p2a451/arm64-v8a \
        appmodules \
        react_codegen_rngesturehandler_codegen \
        react_codegen_rnscreens \
        react_codegen_safeareacontext
    from /Users/nikitashanin/projects/react-native-safe-area-context/example/android/app
  CMake Warning at /Users/nikitashanin/Library/Android/sdk/ndk/23.1.7779620/build/cmake/android-legacy.toolchain.cmake:416 (message):
    An old version of CMake is being used that cannot automatically detect
    compiler attributes.  Compiler identification is being bypassed.  Some
    values may be wrong or missing.  Update to CMake 3.19 or newer to use
    CMake's built-in compiler identification.
  Call Stack (most recent call first):
    /Users/nikitashanin/Library/Android/sdk/ndk/23.1.7779620/build/cmake/android.toolchain.cmake:55 (include)
    /Users/nikitashanin/projects/react-native-safe-area-context/example/android/app/.cxx/Debug/69p2a451/arm64-v8a/CMakeFiles/3.18.1-g262b901/CMakeSystem.cmake:6 (include)
    CMakeLists.txt:28 (project)
    ```

Not sure what goes wrong. Unless I am really supposed to make changes to that cpp code myself.

File: ios/RNCSafeAreaViewEdgeModes.m (hunk 7 above)
Comment (Shaninnik) [reply ios/RNCSafeAreaViewEdgeModes.m]: Yeah I know, but I just can't get it to work with RCTConvert. It is killing me. Having `RNCSafeAreaViewEdgeModeFromString` was the only way it worked.  I can get rid of 

```
RCT_ENUM_CONVERTER(
    RNCSafeAreaViewEdgeMode,
    (@{
      @"off" : @(RNCSafeAreaViewEdgeModeOff),
      @"additive" : @(RNCSafeAreaViewEdgeModeAdditive),
      @"maximum" : @(RNCSafeAreaViewEdgeModeMaximum),
    }),
    RNCSafeAreaViewEdgeModeOff,
    integerValue);
    
```
    
 Since that `RNCSafeAreaViewEdgeModeFromString` does the same thing int RCT_CUSTOM_CONVERTER for `RNCSafeAreaViewEdges` and we never pass `RNCSafeAreaViewEdgeMode` from JS directly

File: src/specs/NativeSafeAreaView.ts (hunk 10 above)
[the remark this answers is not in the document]
Comment (jacobp100) [reply src/specs/NativeSafeAreaView.ts]: This is only the native code, so you could just do `readonly string[]`, and then always give an array with exactly 4 values corresponding to `top`, `right`, `bottom`, `left`. Less clean, but we know that an array of strings does work with fabric

File: ios/RNCSafeAreaViewEdgeMode.h [hunk 11]
```diff
@@ -0,0 +1,11 @@
+#import <Foundation/Foundation.h>
+
+typedef NS_ENUM(NSInteger, RNCSafeAreaViewEdgeMode) {
+  RNCSafeAreaViewEdgeModeOff,
+  RNCSafeAreaViewEdgeModeAdditive,
+  RNCSafeAreaViewEdgeModeMaximum
+};
+
+@interface NSString (EnumParser)
```
jacobp100, line 9: Add this, then if you import this header into `RNCSafeAreaViewEdges.m`, you'll be able to use `RCT_ENUM_CONVERTER`

```suggestion
#import <React/RCTConvert.h>

@interface RCTConvert (RNCSafeAreaViewEdgeMode)
+ (RNCSafeAreaViewEdgeMode)RNCSafeAreaViewEdgeMode:(nullable id)json;
@end
```

ios/RNCSafeAreaViewEdgeModes.m
jacobp100: Added a comment I _think_ is the issue you're facing

src/specs/NativeSafeAreaView.ts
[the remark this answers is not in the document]
Shaninnik: Nope can't do. `edges` is not an array of string anymore, it is an object. If I change it here back to `readonly string[];` then Android will not build even for old arch, it will demand `setEdges(view: SafeAreaView!, value: ReadableArray?)` in `SafeAreaViewManager.kt`

I am looking at it

Shaninnik:
@jacobp100 fabric works now, both iOS and Android

src/SafeAreaView.tsx
```diff
@@ -27,7 +29,20 @@ export const SafeAreaView = React.forwardRef<
           return accum;
         }, {})
       : edges;
-    return _edges ?? defaultEdges;
+
```
jacobp100, line 32: Minor - but we convert the array to an object, then use the object to create a new object. We could

android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaView.kt
```diff
@@ -96,11 +96,16 @@ class SafeAreaView(context: Context?) :
     updateInsets()
   }
 
-  fun setEdges(edges: EnumSet<SafeAreaViewEdges>?) {
+  fun setEdges(edges: SafeAreaViewEdges) {
     mEdges = edges
     updateInsets()
   }
 
+//  fun setMinPadding(minPadding: EdgeInsets?) {
```
jacobp100, line 104: Do you still need this?

Shaninnik: oops, leftovers, cleaned it

example/android/gradle.properties
```diff
@@ -37,7 +37,7 @@ reactNativeArchitectures=armeabi-v7a,arm64-v8a,x86,x86_64
 # your application. You should enable this flag either if you want
 # to write custom TurboModules/Fabric components OR use libraries that
 # are providing them.
-newArchEnabled=false
+newArchEnabled=true
```
jacobp100, line 40: Will need to revert this before we merge

ios/RNCSafeAreaShadowView.m
```diff
@@ -109,23 +110,29 @@ - (void)updateInsets
...
-    super.marginBottom = (YGValue){insetBottom + bottom, YGUnitPoint};
-    super.marginLeft = (YGValue){insetLeft + left, YGUnitPoint};
+    super.marginTop = (YGValue){[self getEdgeValue:edges.top insetValue:insets.top edgeValue:top], YGUnitPoint};
+    super.marginRight = (YGValue){[self getEdgeValue:edges.right insetValue:insets.right edgeValue:right], YGUnitPoint};
+    super.marginBottom = (YGValue){[self getEdgeValue:edges.bottom insetValue:insets.bottom edgeValue:bottom], YGUnitPoint};
+    super.marginLeft = (YGValue){[self getEdgeValue:edges.left insetValue:insets.left edgeValue:left], YGUnitPoint};
+  }
+}
+
+- (CGFloat)getEdgeValue:(RNCSafeAreaViewEdgeMode) edgeMode insetValue:(CGFloat)insetValue edgeValue:(CGFloat) edgeValue
```
jacobp100, line 128:
```suggestion
- (CGFloat)getEdgeValue:(RNCSafeAreaViewEdgeMode)edgeMode insetValue:(CGFloat)insetValue edgeValue:(CGFloat)edgeValue
```

sorry I’m being fussy 🤣

src/SafeAreaView.tsx
jacobp100: (Not essential to change though)

ios/RNCSafeAreaShadowView.m
```diff
@@ -6,6 +6,7 @@
 #import "RNCSafeAreaViewEdges.h"
```
jacobp100, line 6: Is this file used anymore? We might be able to delete it

Shaninnik: You mean delete header import? RNCSafeAreaViewEdges.h is where RNCSafeAreaViewEdges struct lives

jacobp100 (approved):

ios/RNCSafeAreaShadowView.m
jacobp100: https://github.com/th3rdwave/react-native-safe-area-context/blob/main/ios/RNCSafeAreaViewEdges.h and https://github.com/th3rdwave/react-native-safe-area-context/blob/main/ios/RNCSafeAreaViewEdges.m - does these still get used?

jacobp100: Sorry - I see what you mean. Maybe we could rename it RNCSafeAreaViewEdgeMode - so it matches the type name

Shaninnik:
I've pushed a readme update. I am bad at explaining and english is not my native language, feel free to suggest edits

jacobp100:
Yep! Hopefully shouldn't be too long!

janicduplessis:
Looks good, thanks @Shaninnik @jacobp100

janicduplessis:
I will merge this and do a few minor adjustments

jacobp100:
@Shaninnik Thanks for your work on this! It's in the latest release

Shaninnik:
Thanks @janicduplessis @jacobp100 appreciate your comments and help on this!

PR #407, opened 2023-07-03
Update safe area frame when keyboard shows/hides

jacobp100:
When the keyboard shows/hides in pageSheet/formSheet presentation styles, the modal is pushed partially up, so the `y` position changes

### DIFF
--- a/ios/Fabric/RNCSafeAreaProviderComponentView.mm
+++ b/ios/Fabric/RNCSafeAreaProviderComponentView.mm
@@ -24,6 +24,15 @@ using namespace facebook::react;
   if (self = [super initWithFrame:frame]) {
     static const auto defaultProps = std::make_shared<const RNCSafeAreaProviderProps>();
     _props = defaultProps;
+
+    [NSNotificationCenter.defaultCenter addObserver:self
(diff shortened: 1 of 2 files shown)


janicduplessis:
Do we need to cleanup the observer when the view is removed?

jacobp100:
Technically no - but I can add it if you’d feel more comfortable

https://developer.apple.com/documentation/foundation/notificationcenter/1413994-removeobserver

> If your app targets iOS 9.0 and later or macOS 10.11 and later, and you used [addObserver(_:selector:name:object:)](https://developer.apple.com/documentation/foundation/notificationcenter/1415360-addobserver), you do not need to unregister the observer. If you forget or are unable to remove the observer, the system cleans up the next time it would have posted to it.

jacobp100:
That said - it might be necessary to do more granular for fabric, which recycles components. Do you know where that logic is meant to go? Maybe in `didMoveToWindow`, and either subscribe when the window is non-nil, then unsubscribe when the window is nil?

janicduplessis (approved):

janicduplessis:
I think this is fine

Issue #364, opened 2023-03-09
SafeAreaView flickers with incorrect initial insets


janicduplessis:
This seems to happen with translucent status bar, when it is initially rendered on app start.

jacobp100:
Can you provide a minimal repro?

janicduplessis:
@jacobp100 I discovered this issue in the expensify app in the PR https://github.com/Expensify/App/pull/15778, will work on getting a minimal repro in the example app soon.

janicduplessis:
I think they are, but wrongly have a bottom value, which might be what causes the flicker. I noticed by logging initialWindowMetrics. Might be related to the flags the app window has initially.

jacobp100:
Oh interesting. I wonder if android does the same thing of reporting only the safe area that overlaps with the view itself

todorone:
It may be connected with this issue - https://github.com/th3rdwave/react-native-safe-area-context/issues/186, which @jacobp100 by some reason closed without explanation on how it could be solved or workarounded.

jacobp100:
I closed it because there’s not enough information to reproduce it

zabojad:
I'm seing the same behavior with `"react-native-safe-area-context": "4.2.5"` on Android only...

jacobp100:
You need a full repo showing it because the config you put in your android manifest affects the behaviour of the library

zabojad:
@jacobp100 Interesting! What exact parameters from the Android manifest file do affect the behavior of this lib?

jacobp100:
I don't work on Android anymore - so I don't remember

We have a ticket for this https://github.com/th3rdwave/react-native-safe-area-context/issues/349 - it would be really useful if someone is able to help out with this - because I won't be able to

HenrikZabel:
@janicduplessis did you manage to fix it? If yes, could you share how you did it?

toviszsolt:
Hello guys!

- Here is an example video, slow down to see what is happening.
- Layout shift happens on the first and fourth start, there are 4 starts in total.
- I was never able to replicate the phenomenon in any way at Expo Go.
- This effect only occurs in final live builds. I used EAS Build / Play Console.
- I also noticed that if I clear the application data (not the cache, but everything), the first startup is always good.
- The occurrence of the problem seems to be completely random, sometimes 10 times in a row.
- @jacobp100  I can give you access to my private repo so you can have test code. There is only one not too complicated screen. Just please let me know if you need it.

[...]

Issue #234, opened 2022-01-28
Translucent StatusBar causes flicker on render heavy screens


vasafix:
Translucent StatusBar causes useSafeAreaInsets to return incorect values causing flicker

reprepo with bug example:
https://github.com/vasafix/translucent-status-bar-bug

video example:
![p2P6](https://user-images.githubusercontent.com/51510519/151564637-4fe3a8f1-c09d-4139-9fde-7a4fade2b054.gif)

react-native: 0.67.1
react-native-safe-area-context: 3.3.2

HenrikZabel:
@vasafix did you manage to fix it? If yes, could you share how you did it?

vasafix:
@HenrikZabel We hardcoded the top inset value, as most Android status bars are about 24px tall, this was not the best solution but was sufficient enough for our use-case.

I Just tested this with `react-native-safe-area-context` version `4.0.1` and it seems, that I am reliably getting top inset value every time, so I guess its time to do this the proper way :)

One major change we did, since I reported this bug, was to use `@react-navigation/stack` instead of `@react-navigation/native-stack`, you could give this a try if you are experiencing this bug even on the latest version of `react-native-safe-area-context`

sujan-york-ie:
worked for me if I try to set the translucent in StatusBar using StatusBar.setTranslucent(true);

PR #422, opened 2023-08-01
Fix invalid area insets on Android

tboba:
## Summary
Hi! 
During the work on `react-native-screens` I found a strange bug in `useSafeAreaInsets` hook on Android side. 
Currently, area insets received by `useSafeAreaInsets` seem to be invalid. On the first render, as the hook is sending proper values (I assume it uses `getSafeAreaInsets` method on getting initial window metrics) it sends incorrect inset (which is 0) after a while.

This Pull Request introduces a fix that adds takes `windowInsets` into comparing values of each inset when the native side reports negative or zero value. It fixes possibly #364 and #234 (if flickering is caused by insets being equal to 0).

<details>
useSafeAreaInsets-Before
<summary>Before</summary>
</details>
<details><summary>After</summary>
useSafeAreaInsets-After
</details>

## Test Plan
1. Wrap your code with `<SafeAreaInsetsContext.Consumer>` and `<SafeAreaProvider>`
2. Try to use `useSafeAreaInsets().top` and then simply `console.log` it.
3. Launch the android emulator.

Before adding the changes it should return `24` on the initial render and `0` on the next render. After adding the changes it should only return `24`, which is expected value.

### DIFF
--- a/android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaUtils.kt
+++ b/android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaUtils.kt
@@ -6,7 +6,6 @@ import android.view.ViewGroup
 import android.view.WindowInsets
 import androidx.annotation.RequiresApi
 import java.lang.IllegalArgumentException
-import kotlin.math.max
 import kotlin.math.min
 
 @RequiresApi(Build.VERSION_CODES.R)
@@ -72,10 +71,18 @@ fun getSafeAreaInsets(view: View): EdgeInsets? {
   val visibleRect = android.graphics.Rect()
   view.getGlobalVisibleRect(visibleRect)
   return EdgeInsets(
-      top = max(windowInsets.top - visibleRect.top, 0f),
-      right = max(min(visibleRect.left + view.width - windowWidth, 0f) + windowInsets.right, 0f),
-      bottom = max(min(visibleRect.top + view.height - windowHeight, 0f) + windowInsets.bottom, 0f),
-      left = max(windowInsets.left - visibleRect.left, 0f))
+      top = maxOf(windowInsets.top - visibleRect.top, windowInsets.top, 0f),
+      right =
+          maxOf(
+              min(visibleRect.left + view.width - windowWidth, 0f) + windowInsets.right,
+              windowInsets.right,
+              0f),
+      bottom =
+          maxOf(
+              min(visibleRect.top + view.height - windowHeight, 0f) + windowInsets.bottom,
+              windowInsets.bottom,
+              0f),
+      left = maxOf(windowInsets.left - visibleRect.left, windowInsets.left, 0f))
 }
 
 fun getFrame(rootView: ViewGroup, view: View): Rect? {
--- a/android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaUtils.kt
+++ b/android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaUtils.kt
@@ -6,7 +6,6 @@ import android.view.ViewGroup
 import android.view.WindowInsets
 import androidx.annotation.RequiresApi
 import java.lang.IllegalArgumentException
-import kotlin.math.min
 
 @RequiresApi(Build.VERSION_CODES.R)
 private fun getRootWindowInsetsCompatR(rootView: View): EdgeInsets? {
@@ -35,7 +34,7 @@ private fun getRootWindowInsetsCompatM(rootView: View): EdgeInsets? {
       // want to be consistent with iOS. Using the min value makes sure we
       // never get the keyboard offset while still working with devices that
       // hide the navigation bar.
-      bottom = min(insets.systemWindowInsetBottom, insets.stableInsetBottom).toFloat(),
+      bottom = minOf(insets.systemWindowInsetBottom, insets.stableInsetBottom).toFloat(),
       left = insets.systemWindowInsetLeft.toFloat())
 }
 
@@ -74,12 +73,12 @@ fun getSafeAreaInsets(view: View): EdgeInsets? {
       top = maxOf(windowInsets.top - visibleRect.top, windowInsets.top, 0f),
       right =
           maxOf(
-              min(visibleRect.left + view.width - windowWidth, 0f) + windowInsets.right,
+              minOf(visibleRect.left + view.width - windowWidth, 0f) + windowInsets.right,
               windowInsets.right,
               0f),
       bottom =
           maxOf(
-              min(visibleRect.top + view.height - windowHeight, 0f) + windowInsets.bottom,
+              minOf(visibleRect.top + view.height - windowHeight, 0f) + windowInsets.bottom,
               windowInsets.bottom,
               0f),
       left = maxOf(windowInsets.left - visibleRect.left, windowInsets.left, 0f))


android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaUtils.kt
```diff
@@ -72,10 +71,18 @@ fun getSafeAreaInsets(view: View): EdgeInsets? {
...
   return EdgeInsets(
-      top = max(windowInsets.top - visibleRect.top, 0f),
-      right = max(min(visibleRect.left + view.width - windowWidth, 0f) + windowInsets.right, 0f),
-      bottom = max(min(visibleRect.top + view.height - windowHeight, 0f) + windowInsets.bottom, 0f),
-      left = max(windowInsets.left - visibleRect.left, 0f))
+      top = maxOf(windowInsets.top - visibleRect.top, windowInsets.top, 0f),
+      right =
+          maxOf(
+              min(visibleRect.left + view.width - windowWidth, 0f) + windowInsets.right,
+              windowInsets.right,
```
jacobp100, line 78:
```suggestion
              min(visibleRect.left + view.width - windowWidth, 0f) + windowInsets.right,
```

Isn't this the same thing, since it's a `min(something, 0) + windowInsets.right`

android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaUtils.kt
```diff
@@ -72,10 +71,18 @@ fun getSafeAreaInsets(view: View): EdgeInsets? {
   val visibleRect = android.graphics.Rect()
   view.getGlobalVisibleRect(visibleRect)
   return EdgeInsets(
-      top = max(windowInsets.top - visibleRect.top, 0f),
-      right = max(min(visibleRect.left + view.width - windowWidth, 0f) + windowInsets.right, 0f),
-      bottom = max(min(visibleRect.top + view.height - windowHeight, 0f) + windowInsets.bottom, 0f),
-      left = max(windowInsets.left - visibleRect.left, 0f))
+      top = maxOf(windowInsets.top - visibleRect.top, windowInsets.top, 0f),
```
jacobp100, line 74: Were you able to verify that `visibleRect` could be less than 0?

android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaUtils.kt
```diff
@@ -72,10 +71,18 @@ fun getSafeAreaInsets(view: View): EdgeInsets? {
...
+      right =
+          maxOf(
+              min(visibleRect.left + view.width - windowWidth, 0f) + windowInsets.right,
```
jacobp100, line 77: If we need to use `maxOf`, could we also use `minOf` for consistency?

android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaUtils.kt
tboba: It's not the same - as I was testing it, the value in `min` was returning negative value (-66, actually) 🤔

android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaUtils.kt
tboba: When I was testing it, the `visibleRect.top` prop was equal to 66 (so the`windowInsets.top - visibleRect.top` was just equal to 0, as 66 was being subtracted from 66). I also think `visibleRect` will never be less than 0, as we're not setting any offset in `getGlobalVisibleRect`, so it will depend only on the width and height, calculated internally.

android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaUtils.kt
tboba: Sure!

tboba: Resolved in https://github.com/th3rdwave/react-native-safe-area-context/pull/422/commits/cbb667ccc9d1b69b9416a1310b23fb5dee16ff3c.

android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaUtils.kt
jacobp100: I'm not an Android developer, but I think the logic here is trying to determine what portion of the view is overlapped by the safe area (which is what iOS does). I think this change is basically removing that logic, and effectively replacing it with `windowInsets`. I think what you need to be ensuring is that your view fully bleeds into the safe area

jacobp100: I don't know if Android has the view hierarchy debugger like Xcode, but it'd be useful if you could share a screenshot. I'd expect the view to be underneath the status bar and the bit at the bottom with the back button etc.

tboba: I looked onto the Layout Inspector (which is something like a view hierarchy debugger in Xcode) and it looks like SafeAreaProvider is already mounted with 24dp, so it already takes status bar and bottom gesture navigator into account.

tboba: So it looks like windowInsets returns values that are being taken from a root view (which matches the dimensions of the window without the status bar and gesture nav). So I'm just wondering, shouldn't it be sufficient to just calculate window insets just from `windowInsets`, if `windowInsets` already gives proper values? 🤔

jacobp100: Thanks for the screenshot. I'm pretty sure that for that layout, you should have zero safe area insets. If that blue selection extended up and down to match the back-most white view, you would have safe areas

jacobp100: Can you follow this and report back?

https://developer.android.com/develop/ui/views/layout/edge-to-edge

I know there's definitely some config changes needed to get non-zero insets on Android - but I don't remember exactly what it was. I'd really like to get to the bottom of this and document it - a lot of people get confused by this

tboba: Sure @jacobp100, I managed to make my status bar and bottom navigation bar transparent aaaaand it looks like I'm getting correct insets 🤔 Strange, so it looks like when the status bar is transparent the inset is positive, which actually makes sense. I think I misunderstood an idea of the safe area on Android 😅 

The changes I've made:
1. In MainActivity I've overrided `onCreate(@Nullable Bundle savedInstanceState)` method, where I added those lines:

```java
WindowCompat.setDecorFitsSystemWindows(this.getWindow(), false);
    WindowInsetsControllerCompat windowInsetsController =
            ViewCompat.getWindowInsetsController(getWindow().getDecorView());
    if (windowInsetsController == null) {
      return;
    }

    windowInsetsController.setAppearanceLightNavigationBars(true);
```

2. In `res/values/styles.xml` file (Basically just a file with declared styles) I've added to main style tag:
```xml
<item name="android:navigationBarColor">
            @android:color/transparent
        </item>

        <!-- Optional: set to transparent if your app is drawing behind the status bar. -->
        <item name="android:statusBarColor">
            @android:color/transparent
        </item>
```
And you should be ready to go 👍 

image

Currently Layout Inspector shows that there is no status bar (as it's transparent).
image

jacobp100: Thanks for confirming! Is any part of this PR needed now?

tboba: I think it's not necessary now, but in case when I will find something related to my issue I will open it 👍 Also, I'm also concerned if this could be just a problem with my emulator? Currently I'm receiving correct values from JS side before applied change 🤔 Anyway, thanks for taking the time on that PR! 🙇

arthurgeron-work:
Why was this closed?

Issue #426, opened 2023-08-29
Why memo onInsetChange then pass non-memo'ed style


Noitidart:
I was thinking of making a PR to improve the memoization and adding imperative get API and also addListener API would that be accepted? The imperative and listener API is useful for things like reanimated.

For instance we wrap onInsetChange with useCallback but then pass style that referential changes every render. What is the purpose for memo'ing onInsetChange? 

https://github.com/th3rdwave/react-native-safe-area-context/blob/930f703ca3337f407e2df13d1f4ec8e9ffcdb654/src/SafeAreaContext.tsx#L89

jacobp100:
You could do two PRs for this

Improving memoization would be useful. Also the use of useCallback there is a bit flaky - we should use the callback function for the `setState`s

I'm not too sure how the addListener would work. If you detail that a bit more I can tell you if there will be any issues

But yes - we do review all PRs 😊

Noitidart:
Thanks very much. I'll do two PR proposato start diacussion and will take it in the direction you recommend.

PR #429, opened 2023-08-29
 Proposal of subscription API - useSafeAreaListener

Noitidart:
## Summary

This is the subscription API I was trying to propose over in #426. Please advise if this is acceptable.

## Test Plan

Currently a proposal, so no tests.

### DIFF
--- a/src/SafeAreaContext.tsx
+++ b/src/SafeAreaContext.tsx
@@ -91,9 +91,22 @@ export function SafeAreaProvider({
     [setFrame, setInsets],
   );
 
+  // Add flex to the style if it doesn't have it.
+  const flexedStyle = React.useMemo(() => {
+    const flatStyle = StyleSheet.flatten(style) || {};
+    if ('flex' in flatStyle === false) {
+      return {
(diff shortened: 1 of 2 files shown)


src/SafeAreaContext.tsx
```diff
@@ -38,61 +51,105 @@ export function SafeAreaProvider({
...
+
+        if (didFrameChange || didInsetsChange) {
+          const nextMetrics = {
+            frame: didFrameChange ? nextFrame : frame,
+            insets: didInsetsChange ? nextInsets : insets,
+          };
+
+          latestMetrics.current = nextMetrics;
+
+          listeners.current.forEach((listener) => listener(nextMetrics));
```
jacobp100, line 106: This isn't a safe thing to do from a set state callback

Noitidart:
> I'm still not really getting what the point of this is. What is this enabling?

I understand. I just shared a reply to this here - https://github.com/th3rdwave/react-native-safe-area-context/pull/428#discussion_r1308745040

PR #431, opened 2023-08-31
Refactor style prop in SafeAreaContext component to avoid unnecessary…

zxcodes:
… re-renders.

In the `SafeAreaContext.tsx` component, the `style` prop was passed using array syntax, which created a new array on each render and caused unnecessary re-renders.

To improve performance, the `style` prop has been refactored to use the spread syntax, which merges the component's default styles with the passed styles object more efficiently. This change ensures that the `SafeAreaContext` component maintains its appearance and behavior while reducing the chance of unnecessary re-renders. The new syntax is `style={{...styles.fill, ...style}}`.

### DIFF
--- a/src/SafeAreaContext.tsx
+++ b/src/SafeAreaContext.tsx
@@ -85,7 +85,7 @@ export function SafeAreaProvider({
 
   return (
     <NativeSafeAreaProvider
-      style={[styles.fill, style]}
+      style={{...styles.fill, ...style}}
       onInsetsChange={onInsetsChange}
       {...others}
     >


jacobp100:
We incidentally have a PR that's doing a similar thing

https://github.com/th3rdwave/react-native-safe-area-context/pull/427/files#diff-7798046c2c2d997663d18e875691ec66f4657a5e2467201b1d76ae1b530a0073R93

This avoids creating an array in the case you don't pass in a style. Since this PR also has some other optimisations, we'll go in that direction

Thanks for the PR anyway 😊

PR #446, opened 2023-10-23
chore: resolve ViewProps type import

jaredspeters:
## Summary

`SafeAreaContext.ts` contains an import for `ViewProps` that does not meet the TypeScript v5 `verbatimModuleSyntax` rule. It is now imported as `type`.

## Test Plan

These changes do not affect testing.

### DIFF
--- a/src/SafeAreaContext.tsx
+++ b/src/SafeAreaContext.tsx
@@ -1,6 +1,5 @@
 import * as React from 'react';
-import { Dimensions, StyleSheet } from 'react-native';
-import type { ViewProps } from 'react-native';
+import { Dimensions, StyleSheet, type ViewProps } from 'react-native';
 import { NativeSafeAreaProvider } from './NativeSafeAreaProvider';
 import type {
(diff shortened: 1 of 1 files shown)


src/SafeAreaContext.tsx
```diff
@@ -1,5 +1,6 @@
 import * as React from 'react';
-import { Dimensions, StyleSheet, ViewProps } from 'react-native';
+import { Dimensions, StyleSheet } from 'react-native';
```
janicduplessis, line 2: Can you add a type qualifier to the existing import instead of adding a 2nd one. Like import { A, type B } from …

janicduplessis (changes requested):

src/SafeAreaContext.tsx
jaredspeters: Done

jacobp100 (approved):

janicduplessis (approved):

PR #454, opened 2023-11-09
React Native 0.73.0 Support

ravindraguptacapgemini:
Fixed issues causing build failure for android on React Native 0.73.0-rc.4


jacobp100 (approved):
LGTM but I think we do have to wait until 0.73 ships before merging, as this will break people on 0.72

janicduplessis:
I think we can publish this as pre-release version. Will have a look in the next few days.

PR #464, opened 2024-01-04
Fix the `pointerEvents` prop on SafeAreaProvider (fixes #432)

### LINKED ISSUE #432: SafeAreaProvider cannot set `pointerEvents` prop
I searched the issues and find one: https://github.com/th3rdwave/react-native-safe-area-context/pull/130

it changes the `SafeAreaView` component so it works fine, but `SafeAreaProvider` stays the same, it doesn't work.

do you mind to take a look at this?

cboar:
## Summary

This PR fixes #432, where the `pointerEvents` prop has no effect when applied to the `SafeAreaProvider`. I share the same use-case as the creator of the issue: I wish to use a `SafeAreaProvider` as part of an overlay, where touch events should be passed to views underneath it, as is expected when setting `pointerEvents="box-none"`.

It seems the behavior described was working until the library was migrated to the new architecture. Here are my findings:
* When running with paper, all props are set via the codegen-created `RNCSafeAreaProviderManagerDelegate#setProperty` method. This method will invoke the appropriate `@ReactProp` setter method defined in its corresponding ViewManager.
* If the `setProperty` method does not include a case to handle a prop, it will not be passed to the ViewManager. Note that react-native _does_ provide cases for most `View` props with the [BaseViewManagerInterface](https://github.com/facebook/react-native/blob/9ba3bc99e496d9e4d246445d63b68b0692b5fcab/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManagerInterface.java#L20). However, `pointerEvents` is not included.
* In order to add the case during codegen, I simply add the `pointerEvents` prop to the `NativeProps` specs definition. Unfortunately, because the definition requires a specific format to pass codegen, it also clashes with the inherited `ViewProps`. I add a `@ts-ignore` comment as a workaround.
* Lastly, I update the `SafeAreaProviderManager` to inherit from `ReactViewManager` to handle the newly defined prop.

## Test Plan

As a test, I modified the example app by styling a `SafeAreaProvider` with `StyleSheet.absoluteFill`, and placing a Touchable button underneath/behind the view. When adding `pointerEvents="none"` to the `SafeAreaProvider`, the button can be pressed with this fix. Without this fix, the button cannot be pressed.

### DIFF
--- a/android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaProviderManager.kt
+++ b/android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaProviderManager.kt
@@ -4,28 +4,29 @@ import com.facebook.react.bridge.ReactContext
 import com.facebook.react.module.annotations.ReactModule
 import com.facebook.react.uimanager.ThemedReactContext
 import com.facebook.react.uimanager.UIManagerHelper
-import com.facebook.react.uimanager.ViewGroupManager
 import com.facebook.react.viewmanagers.RNCSafeAreaProviderManagerDelegate
 import com.facebook.react.viewmanagers.RNCSafeAreaProviderManagerInterface
+import com.facebook.react.views.view.ReactViewGroup
+import com.facebook.react.views.view.ReactViewManager
 
 @ReactModule(name = SafeAreaProviderManager.REACT_CLASS)
 class SafeAreaProviderManager :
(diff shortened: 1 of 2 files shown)


jacobp100 (approved):
Makes sense. Thanks!

jacobp100:
@janicduplessis is this ok to merge?

janicduplessis:
Sorry about taking forever to get to this, I wonder if that is something that would be better fixed in RN core. Pointer events should be part of the props that are inherited from View, and it does work on iOS.

PR #473, opened 2024-02-22
fix: fixes for new arch on RN 74

WoLewicki:
## Summary

This PR aligns the code with newest version of `yoga` in RN `0.74.0-rc.0`.  Made with huge help from @piaskowyk

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -41,8 +41,8 @@ jobs:
       - name: Install example app dependencies
         run: yarn install
         working-directory: example
-      - name: Build android example app
-        run: ./gradlew assembleDebug
+      - name: Build android example app with new arch disabled
+        run: ./gradlew assembleDebug -PnewArchEnabled=false
         working-directory: example/android
   android-build-fabric:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


jacobp100:
Thanks for this! What does this do for older versions? Will they break?

WoLewicki:
On new arch it will for sure. On paper I think those files are not compiled so it should not.

jacobp100:
I think last time we published this as a pre-release, and then waited until it fully drops to publish in full

@janicduplessis

janicduplessis:
Yes, we can do pre-release, I'm fine breaking back compat for new arch still.

WoLewicki:
So let's do it please, it will help with testing it in other libs which use SAC 🚀

janicduplessis (approved):

janicduplessis:
Looks like this will cause issues even with new arch disabled. The ci build on android fails with old arch and rn 0.73. We will need to find a way to make this backwards compatible. Maybe we can just not compile those classes with some preprocessor macros when on old arch.

WoLewicki:
Yeah, it seems not right that you are trying to compile shadow nodes on old arch.

WoLewicki:
Ok it seems that you are building new arch in both flows now: https://github.com/th3rdwave/react-native-safe-area-context/blob/c9b729343e8e2a73b723a76bae6319b97b121cc6/example/android/gradle.properties#L37 😄

PR #474, opened 2024-02-28
Fix no rendered content on CarPlay

### LINKED ISSUE #135: SafeAreaProvider doesn't render content  if the app is launched from CarPlay
If the app is launched from carplay, the insets here will be null which cause SafeAreaProvider doesn't render content
https://github.com/th3rdwave/react-native-safe-area-context/blob/b1824ef84d78a44e91cd156a29abe3da243eee94/src/SafeAreaContext.tsx#L78

tommynordli:
## Summary

When having the app opened on CarPlay only, there are no safe area insets, resulting in the provider just returning `null`. Just returning the children instead does the trick. This should fix #135

### DIFF
--- a/src/SafeAreaContext.tsx
+++ b/src/SafeAreaContext.tsx
@@ -100,7 +100,7 @@ export function SafeAreaProvider({
             {children}
           </SafeAreaInsetsContext.Provider>
         </SafeAreaFrameContext.Provider>
-      ) : null}
+      ) : children}
     </NativeSafeAreaProvider>
   );
 }


tommynordli:
I didn't spend too much time digging into it so I'm not sure exactly why, but I don't think safearea insets on CaPlay aren't really neccessary as it's mostly template based. This only happens when launching the app from CarPlay with the app closed on the device btw, not if the app is initialized on the phone first.

jacobp100:
We can't merge this change because it'll cause excessive renders. If you figure out why the insets aren't getting set, we can look at merging something that fixes that issue

janicduplessis:
Yea, we need to keep the current behaviour of rendering null until we receive the first inset events. Is there a way to know it is carplay from JS? Or we could manually send an insets event with 0 insets if carplay from native code.

jacobp100:
Closing - if you find the root cause, feel free to ping me or open a new PR

mefjuu:
@janicduplessis Have you managed to obtain the flag while JS bundle is running in carplay?

jacobp100:
@mefjuu nobody is working on this. Feel free to take it on if you wish

PR #479, opened 2024-03-13
update build.gradle to fix #478

### LINKED ISSUE #478: Add support for react native 0.73 which uses android gradle version of 8.1.1 which comes pre bundled with react native
We get following error when compiling this library for react native 0.73+ versions : 

```
The project is using an incompatible version (AGP 8.1.1) of the Android Gradle plugin. Latest supported version is AGP 7.3.1
```

In your build.gradle file , remove the version u are using and let the project pick the version that uses your library. will raise a PR for this

nateshmbhat:
## Summary

This removes the hardcoded gradle plugin version (7.3.1) so that it uses the gradle plugin from the root project that uses this library.

## Test Plan
Just include this library in a react native app that uses 0.73+ version of react native and run the app

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -8,7 +8,7 @@ buildscript {
 
     dependencies {
         classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version")
-        classpath("com.android.tools.build:gradle:7.3.1")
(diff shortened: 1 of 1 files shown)


nateshmbhat:
looks like the issue was due to older version of android studio :/
closing this

PR #480, opened 2024-03-13
fix: added run recalculation for safe_area when window did become visible

ruslan-deyarov:
Added run recalculation for safe_area when window did become visible

## Summary

We found missed scenario when we have to run recalculation for safe_area. In case we open before native modal and then   put our app into background mode. After restore it from background and close this modal we see old values for safe_area. We need to run recalculation to cover this scenario.

Native modal partially cover screen and in this case safe_area is reduced a little bit by iOS.

Screenshot 2024-03-08 at 15 32 09

### DIFF
--- a/ios/RNCSafeAreaProvider.m
+++ b/ios/RNCSafeAreaProvider.m
@@ -26,6 +26,10 @@
                                            selector:@selector(invalidateSafeAreaInsets)
                                                name:UIKeyboardDidChangeFrameNotification
                                              object:nil];
+    [NSNotificationCenter.defaultCenter addObserver:self
+                                           selector:@selector(invalidateSafeAreaInsets)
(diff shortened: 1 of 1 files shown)


jacobp100:
Do you have a reproduction here?

ruslan-deyarov:
> Do you have a reproduction here?

Yes, let me attach video.

ruslan-deyarov:
Wrong safe_area size after background and closing native modal on top.

https://github.com/th3rdwave/react-native-safe-area-context/assets/109741896/c0d5c2fa-9aaa-405b-92cb-eb4625b25d9d

jacobp100 (approved):

janicduplessis:
Thanks, sorry about the delay merging this!

PR #489, opened 2024-04-11
Prevent from crashing with null cannot be cast to non-null type `kotlin.Any`

lukmccall:
## Summary

It's possible for the `getInitialWindowMetrics` method to return `null`, which can cause crashes if we force cast it to `Any`. I faced issues with reloading the app, resulting in errors:
```
ERROR  Error: Exception in HostObject::get for prop 'RNCSafeAreaContext': java.lang.NullPointerException: null cannot be cast to non-null type kotlin.Any, js engine: hermes
```

I'm unsure of the expected behavior when `getInitialWindowMetrics` returns null, but the current exception doesn't provide enough information.

## Test Plan

- Tested in expo app ✅


janicduplessis (approved):

janicduplessis:
Looks good, thanks!

PR #507, opened 2024-06-28
fix: use kotlin compatible getType

WoLewicki:
## Summary

PR adding compatibility for RN `0.75.0-rc.2`. Since `Dynamic` class has been migrated to Kotlin: https://github.com/facebook/react-native/blob/e320ab47cf855f2e5de74ea448ec292cf0bbb29a/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/bridge/Dynamic.kt, the name of function to use is strictly `getType`. It is compatible with previous versions so should be safe to just merge.


janicduplessis (approved):

janicduplessis:
Thanks!

PR #535, opened 2024-10-11
[android] Remove a check that is no longer needed

chrfalch:
## Summary

We've had some issue with monorepos when resolving the codegen config on Android in other libraries, and found that due to versions no longer in use we could now remove the version test in `react-native-config.js` and always return the component descriptor etc.

(The monorepo issue was with using `require.main.require` which will resolve to the calling script and not the module)

For reference, here is the same PR in `@react-native-community/slider`:

https://github.com/callstack/react-native-slider/pull/657

## Test Plan

This will not fix a bug as of now, but is a preparation for later to make sure that the component descriptors are correctly added to the `autolinking.cpp` file.

In `@react-native-community/slider` the following issue was registered: https://github.com/callstack/react-native-slider/issues/652


jacobp100:
Thanks for this. I'm not too familiar with Android. Janic might know. Otherwise, we might just wait and see what CallStack do with the other PR?

janicduplessis:
I am a bit worried about backwards compat here, could we add a fallback to something like require("../@react-native-community/…") to handle that case? Assuming deps are in the same folder. I would also be open to just setting the default to true so if we cant find the file we assume codegenconfig is supported.

chrfalch:
> I am a bit worried about backwards compat here, could we add a fallback to something like require("../@react-native-community/…") to handle that case? Assuming deps are in the same folder. I would also be open to just setting the default to true so if we cant find the file we assume codegenconfig is supported.

Understand your concerns - but would codegen even be used in an older version of React Native? We could also just change `require.main.require` -> `require` - which also works.

chrfalch:
The corresponding PR was merged in `@react-native-community/slider` now.

janicduplessis:
The problem that this code fixes is that older versions of RN cli do not support the codegen config and throw an error because of that extra config prop sadly.

chrfalch:
> If we want to keep this simple I would be fine just changing the default value so if the file is not found we add the codegen config.

Reverted and added a change to the initial value only as you suggested :)

vonovak (approved):

PR #568, opened 2024-12-30
fix: Fix RN 78 compatibility (`unit()` was removed)

mrousavy:


jacobp100:
Is this backwards compatible?

mrousavy:
Good question. Probably not, might require a `#if` RN version check.

janicduplessis:
Yea, I think we will have to start to commit to some backward compat for new arch since it is now enabled by default.

I think ifdef would be good, but might need to change build files to pass the RN version.

mrousavy:
oh lol i forgot i already opened this pr hahah my bad

PR #577, opened 2025-01-21
feat: add component provider field for codegen

fobos531:
## Summary

```[Codegen] [DEPRECATED] react-native-safe-area-context should add the 'ios.componentProvider' property in their codegenConfig```

See: https://x.com/o_kwasniewski/status/1876960622832697391

CleanShot 2025-01-21 at 17 31 27@2x

## Test Plan

RNTA works as expected - haven't noticed any regressions

### DIFF
--- a/package.json
+++ b/package.json
@@ -135,6 +135,12 @@
     "android": {
       "javaPackageName": "com.th3rdwave.safeareacontext"
     },
+    "ios": {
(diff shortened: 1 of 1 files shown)


janicduplessis:
Looks good, thanks!

janicduplessis (approved):

PR #578, opened 2025-01-23
feat: Add react-native 0.78 support

mrousavy:
## Summary
In react-native 0.78, they removed `StyleLength::unit()`. Instead, you can check `StyleLength::isUndefined()` directly, which achieves the same thing.

Technically not backwards compatible, but I remember Janic saying the policy for new arch is to not be backwards compatible until it is stable and used by (pretty much) everyone.

## Test Plan
Build an app with RN 78. Currently in RC.1


janicduplessis:
I think it is fine to not be backwards compatible, but will have to release this as beta until 0.78 is stable.

janicduplessis (approved):

PR #589, opened 2025-02-18
chore: suppress `-Wdollar-in-identifier-extension` compiler warning

kkafar:
## Summary

Following https://github.com/software-mansion/react-native-screens/pull/2707

Variable names prefixed with `$` are heavily utilised in codegened code. This leads to a wall of warnings popping up during every Android build. This PR aims to suppress these warnings.

image

☝🏻 and that is not all...

## Test Plan

Just build Fabric + Android with this change.


janicduplessis (approved):
Thanks @kkafar !

PR #590, opened 2025-02-21
Fix issues with inset calculation for apps targeting Android SDK 35+

clangen:
## Summary

We recently ran into an issue with `react-native-safe-area-context` returning `0` for top and bottom insets after upgrading our project to Android SDK 35. Turns out others have experienced this issue as well:
* https://github.com/AppAndFlow/react-native-safe-area-context/issues/552
* https://github.com/AppAndFlow/react-native-safe-area-context/issues/546

We addressed this issue by changing the strategy used to resolve the insets; instead of using `ViewTreeObserver.OnPreDrawListener` to detect changes, this commit migrates to using `ViewCompat.setOnApplyWindowInsets`, which seems to be the preferred way to calculate insets anyway.

**Note**: this is against the 4.x branch, as we are not able to use 5.x yet, although the changes should still be applicable. 

## Test Plan

N/A

### DIFF
--- a/android/src/main/java/com/th3rdwave/safeareacontext/EdgeInsets.kt
+++ b/android/src/main/java/com/th3rdwave/safeareacontext/EdgeInsets.kt
@@ -1,3 +1,7 @@
 package com.th3rdwave.safeareacontext
 
-data class EdgeInsets(val top: Float, val right: Float, val bottom: Float, val left: Float)
+data class EdgeInsets(val top: Float, val right: Float, val bottom: Float, val left: Float) {
+    override fun toString(): String {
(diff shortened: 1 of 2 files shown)


janicduplessis:
Hello @clangen, thanks for the PR! I will have a look next week. In the meantime could you just remove any leftover debug logs.

clangen:
After testing across a wider range of devices this patch does not appear to work comprehensively. I'm going to close the MR for now to avoid confusion. If/when we can figure out a better patch I'll re-open.

janicduplessis:
Thanks, I will also try to investigate on my end next week.

PR #591, opened 2025-02-25
feat: add new arch support to macos

gabrieldonadel:
## Summary

Follow up of https://github.com/AppAndFlow/react-native-safe-area-context/pull/513

While building BareExpo for macOS I noticed that macOS builds are broken when using the new arch. This PR added the necessary platform checks to fix it 

## Test Plan

Manually added macOS to the sample app so I could test it manually.

### DIFF
--- a/ios/Fabric/RNCSafeAreaViewComponentView.mm
+++ b/ios/Fabric/RNCSafeAreaViewComponentView.mm
@@ -58,10 +58,10 @@ using namespace facebook::react;
                                                                           _providerView.safeAreaInsets.bottom,
                                                                           _providerView.safeAreaInsets.right];
   NSString *currentSafeAreaInsetsString = [NSString stringWithFormat:@"{%f,%f,%f,%f}",
-                                                                    _currentSafeAreaInsets.top,
-                                                                    _currentSafeAreaInsets.left,
-                                                                    _currentSafeAreaInsets.bottom,
-                                                                    _currentSafeAreaInsets.right];
+                                                                     _currentSafeAreaInsets.top,
+                                                                     _currentSafeAreaInsets.left,
+                                                                     _currentSafeAreaInsets.bottom,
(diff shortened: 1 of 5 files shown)


ios/Fabric/RNCSafeAreaViewComponentView.mm
```diff
@@ -47,11 +47,34 @@ - (NSString *)description
...
     superDescription = [superDescription substringToIndex:superDescription.length - 1];
   }
+  
+#if TARGET_OS_IPHONE
+  NSString *providerViewSafeAreaInsetsString = NSStringFromUIEdgeInsets(_providerView.safeAreaInsets);
+  NSString *currentSafeAreaInsetsString = NSStringFromUIEdgeInsets(_currentSafeAreaInsets);
+#elif TARGET_OS_OSX
+  NSString *providerViewSafeAreaInsetsString;
+  NSString *currentSafeAreaInsetsString;
+  if (@available(macOS 11.0, *)) {
```
jacobp100, line 57: Does RN actually support this target?

gabrieldonadel: This is required if we want to support 0.76, from 0.77 onwards the minimum required version for macOS is 11.0. Should I remove this check?

jacobp100: We only support new arch for the latest versions of RN, so sounds like we could just remove this - or remove it to the extent the compiler will allow

janicduplessis:
I think this is fine, thanks everyone!

janicduplessis:
Can you just run yarn format:write to fix formatting lint.

janicduplessis (approved):

gabrieldonadel:
> Can you just run yarn format:write to fix formatting lint.

Done

Issue #485, opened 2024-03-27
insets.top switching to 0 during iPad orientation change 


Gregoirevda:
On iPad, `insets.top` changes during an orientation change from X to 0 and back to X, causing the app to re-render and recalculate layouts back and forth.

This only happens when going from landscape to portrait.
I've opened a related issue on RN-navigation: https://github.com/react-navigation/react-navigation/issues/11915

jacobp100:
Does it happen with just this library - or is it only with react navigation?

Gregoirevda:
@jacobp100 
I've tried and happens just with react-native-safe-area-context

```jsx
import { SafeAreaProvider, useSafeAreaInsets } from 'react-native-safe-area-context';

export default function App() {
  return (
    <SafeAreaProvider>
      <Kid />
    </SafeAreaProvider>
  );
}

function Kid() {
  const insets = useSafeAreaInsets();
  console.log({ insets });

  return null;
}
```

```logs
// Opening the app on iPad in portrait mode
[2]  LOG  {"insets": {"bottom": 20, "left": 0, "right": 0, "top": 74}}

// Changing orientation to landscape (switches to 0 and back to 20)
[2]  LOG  {"insets": {"bottom": 0, "left": 0, "right": 0, "top": 74}}
[2]  LOG  {"insets": {"bottom": 20, "left": 0, "right": 0, "top": 74}}
// End orientation change

// Changing orientation back to portrait
[2]  LOG  {"insets": {"bottom": 0, "left": 0, "right": 0, "top": 74}}
[2]  LOG  {"insets": {"bottom": 0, "left": 0, "right": 0, "top": 0}}
[2]  LOG  {"insets": {"bottom": 0, "left": 0, "right": 0, "top": 74}}
[2]  LOG  {"insets": {"bottom": 20, "left": 0, "right": 0, "top": 74}}
// End orientation change
```

As all iOS apps with a blurred header put `<ScrollView contentContainerStyle={{paddingTop: insets.top}}` 
and all apps having a blurred bottom tab navigation have `<ScrollView contentContainerStyle={{paddingBottom: insets.bottom}}`
This causes the ScrollView to lag a lot when changing orientation

jacobp100:
[...]

Issue #172, opened 2021-01-26
useSafeAreaInsets emit too many results when iOS rotation


mym0404:
Because of the device performance issue, the repeated emission of the same insets or useless emission from `useSafeAreaInsets` should be skipped.

When I rotated my device with `iPhone X`, the console printed out like the following.

```
[Tue Jan 26 2021 11:40:03.498]  LOG      openFullScreen
[Tue Jan 26 2021 11:40:03.512]  LOG      {"bottom": 34, "left": 0, "right": 0, "top": 44}
[Tue Jan 26 2021 11:40:04.220]  LOG      {"bottom": 21, "left": 44, "right": 0, "top": 0}
[Tue Jan 26 2021 11:40:04.267]  LOG      {"bottom": 21, "left": 44, "right": 0, "top": 0}
[Tue Jan 26 2021 11:40:04.644]  LOG      {"bottom": 21, "left": 44, "right": 0, "top": 0}
[Tue Jan 26 2021 11:40:04.930]  LOG      {"bottom": 0, "left": 0, "right": 0, "top": 0}
[Tue Jan 26 2021 11:40:05.251]  LOG      {"bottom": 0, "left": 44, "right": 0, "top": 0}
[Tue Jan 26 2021 11:40:05.558]  LOG      {"bottom": 21, "left": 44, "right": 44, "top": 0}
[Tue Jan 26 2021 11:40:05.827]  LOG      {"bottom": 21, "left": 44, "right": 44, "top": 0}
```

```
[Tue Jan 26 2021 11:40:08.221]  LOG      closeFullScreen
[Tue Jan 26 2021 11:40:08.325]  LOG      {"bottom": 21, "left": 44, "right": 44, "top": 0}
[Tue Jan 26 2021 11:40:08.491]  LOG      {"bottom": 0, "left": 0, "right": 0, "top": 44}
[Tue Jan 26 2021 11:40:08.793]  LOG      {"bottom": 0, "left": 0, "right": 0, "top": 44}
[Tue Jan 26 2021 11:40:09.234]  LOG      {"bottom": 0, "left": 0, "right": 0, "top": 44}
[Tue Jan 26 2021 11:40:09.394]  LOG      {"bottom": 0, "left": 0, "right": 0, "top": 0}
[Tue Jan 26 2021 11:40:09.677]  LOG      {"bottom": 0, "left": 0, "right": 0, "top": 44}
[...]

PR #592, opened 2025-02-26
feat: coalesced event

### LINKED ISSUE #579: (Android) Insets.top changing from 0 to X and X to 0 loop
## Environment
react-native-safe-area-context: 5.1.0
expo: 52
react-native: 0.76.6

## Issue

I'm having an issue on Android with the insets.top making the screen flicker. This is due to the value changing from 0 to X and X to 0 non-stop.

## Stacktrace

I have this code in a component:

```
const insets = useSafeAreaInsets();
console.log(insets);
```

and by running this in the Android emulator, I get something like this:

```
...
 (NOBRIDGE) LOG  {"bottom": 0, "left": 0, "right": 0, "top": 0}
 (NOBRIDGE) LOG  {"bottom": 0, "left": 0, "right": 0, "top": 24}
 (NOBRIDGE) LOG  {"bottom": 0, "left": 0, "right": 0, "top": 0}
 (NOBRIDGE) LOG  {"bottom": 0, "left": 0, "right": 0, "top": 24}
 (NOBRIDGE) LOG  {"bottom": 0, "left": 0, "right": 0, "top": 0}
 (NOBRIDGE) LOG  {"bottom": 0, "left": 0, "right": 0, "top": 24}
 (NOBRIDGE) LOG  {"bottom": 0, "left": 0, "right": 0, "top": 0}
 (NOBRIDGE) LOG  {"bottom": 0, "left": 0, "right": 0, "top": 24}
 (NOBRIDGE) LOG  {"bottom": 0, "left": 0, "right": 0, "top": 0}
 (NOBRIDGE) LOG  {"bottom": 0, "left": 0, "right": 0, "top": 24}
...
```

Gregoirevda:
## Summary
`onInsetsChange` can sometimes be sent very quickly after another. Using the same technique as the ScrollView, we keep the view property `onInsetsChange`, but it's triggered from a custom `RCTEvent` with a coalescingKey.

If events are sent 5 times very quickly, only 1 event will be sent with the latest insets and frame.

Fixes #579 #485 #172 

Note:
I've tested this in a real app and `onInsetsChange` calls goes from 3 to 1, but in the example app it doesn't change (still 3 calls when rotating) and that's because it doesn't change anything in batching which is good. 
So, if during a flush to JS 5 `onInsetsChange` are sent only the last one will survive, but if the app isn't busy and it can send the 5 events independently it will.

`EventEmitter` will collect the event in a queue and then flush them. Coalesce changes a flush with multiple events from
`[onInsetsChange x1 y1, onInsetsChange x2 y2]` to
`[onInsetsChange x2 y2]`

PS:
I've only implemented the old architecte, if this can be reviewed I can then only apply changes to this part and finish the fabric arch.
Requires `RCT_NEW_ARCH_ENABLED=0 pod install` to test.

### DIFF
--- a/ios/RNCSafeAreaProvider.m
+++ b/ios/RNCSafeAreaProvider.m
@@ -84,7 +84,6 @@
       CGRectEqualToRect(frame, _currentFrame)) {
     return;
   }
-  
   _initialInsetsSent = YES;
   _currentSafeAreaInsets = safeAreaInsets;
   _currentFrame = frame;
--- a/ios/RNCSafeAreaProvider.m
+++ b/ios/RNCSafeAreaProvider.m
@@ -4,7 +4,7 @@
 #import <React/RCTEventDispatcher.h>
 #import <React/RCTUIManager.h>
 #import "RCTUIManagerObserverCoordinator.h"
-#import "RNCChangeEvent.h"
+#import "RNCOnInsetsChangeEvent.h"
 #import "RNCSafeAreaUtils.h"
 
 @interface RNCSafeAreaProvider () <RCTUIManagerObserver>
@@ -89,13 +89,13 @@
 
   [NSNotificationCenter.defaultCenter postNotificationName:RNCSafeAreaDidChange object:self userInfo:nil];
 
-  RNCChangeEvent *changeEvent = [[RNCChangeEvent alloc] initWithEventName:@"onInsetsChange"
+  RNCOnInsetsChangeEvent *onInsetsChangeEvent = [[RNCOnInsetsChangeEvent alloc] initWithEventName:@"onInsetsChange"
                                                                  reactTag:self.reactTag
                                                                    insets:safeAreaInsets
                                                                     frame:frame
                                                             coalescingKey:0];
 
-  [_eventDispatcher sendEvent:changeEvent];
+  [_eventDispatcher sendEvent:onInsetsChangeEvent];
 }
 
 - (void)layoutSubviews
--- a/ios/RNCSafeAreaProvider.m
+++ b/ios/RNCSafeAreaProvider.m
@@ -1,19 +1,33 @@
 #import "RNCSafeAreaProvider.h"
 
+#import <React/RCTEventDispatcher.h>
 #import <React/RCTBridge.h>
 #import <React/RCTUIManager.h>
+#import "RCTUIManagerObserverCoordinator.h"
 #import "RNCSafeAreaUtils.h"
+#import "RNCChangeEvent.h"
+
+@interface RNCSafeAreaProvider () <RCTUIManagerObserver>
+
+@end
 
 @implementation RNCSafeAreaProvider {
+  id<RCTEventDispatcherProtocol> _eventDispatcher;
   UIEdgeInsets _currentSafeAreaInsets;
   CGRect _currentFrame;
   BOOL _initialInsetsSent;
+  uint16_t _coalescingKey;
 }
 
-- (instancetype)init
+- (instancetype)initWithEventDispatcher:(id<RCTEventDispatcherProtocol>)eventDispatcher
 {
-  if ((self = [super init])) {
-#if !TARGET_OS_TV && !TARGET_OS_OSX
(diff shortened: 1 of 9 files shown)


ios/RNCSafeAreaProvider.m
```diff
@@ -69,27 +84,23 @@ - (void)invalidateSafeAreaInsets
       CGRectEqualToRect(frame, _currentFrame)) {
     return;
   }
-
+
```
jacobp100, line 87:
```suggestion

```

ios/RNCSafeAreaProvider.m
```diff
@@ -99,4 +110,13 @@ - (void)layoutSubviews
   [self invalidateSafeAreaInsets];
 }
 
+- (void)dealloc
```
jacobp100, line 113: Minor - could you put this dealloc function after the `init` functions

src/SafeAreaContext.tsx
```diff
@@ -58,6 +58,8 @@ export function SafeAreaProvider({
       nativeEvent: { frame: nextFrame, insets: nextInsets },
     } = event;
 
+    console.log('onInsetsChange', nextInsets);
+
```
jacobp100, line 62:
```suggestion
```

ios/RNCSafeAreaProvider.m
```diff
@@ -1,19 +1,33 @@
...
+@interface RNCSafeAreaProvider () <RCTUIManagerObserver>
+
+@end
 
 @implementation RNCSafeAreaProvider {
+  id<RCTEventDispatcherProtocol> _eventDispatcher;
   UIEdgeInsets _currentSafeAreaInsets;
   CGRect _currentFrame;
   BOOL _initialInsetsSent;
+  uint16_t _coalescingKey;
```
jacobp100, line 19: I can't see how this would ever be changed from the default of `0` - is it actually needed?

ios/RNCSafeAreaProvider.m
```diff
@@ -99,4 +106,12 @@ - (void)layoutSubviews
   [self invalidateSafeAreaInsets];
 }
 
+RCT_NOT_IMPLEMENTED(-(instancetype)initWithFrame : (CGRect)frame)
+RCT_NOT_IMPLEMENTED(-(instancetype)initWithCoder : (NSCoder *)aDecoder)
```
jacobp100, line 110: Minor

```suggestion
RCT_NOT_IMPLEMENTED(-(instancetype)initWithFrame:(CGRect)frame)
RCT_NOT_IMPLEMENTED(-(instancetype)initWithCoder:(NSCoder *)aDecoder)
```

jacobp100 (approved):
Thank you for this!

Interesting it didn't fully resolve the issue. But at least it's not worse

jacobp100:
No - if you add `[_eventDispatcher.bridge.uiManager.observerCoordinator addObserver:self];` and `- (void) uiManagerDidPerformLayout`, the latter function will get called every time RN updates the layout

Gregoirevda:
@jacobp100 I've set a flag in `invalidateSafeAreaInsets` and added the observer in `init` and only emit the event in `uiManagerDidPerformLayout` and this won't work:

- **Initial layout:**
`uiManagerDidPerformLayout` isn't called after the component mounted. I've fixed this by emitting in the `invalidateSafeAreaInsets` when `_initialInsetsSent` hasn't been set.

- **portrait to landscape** event order:
```
uiManagerWillPerform
uiManagerDidPerform
invalidateSafeAreaInsets
```
=> no events sent

- **landscape to portait** event order
```

invalidateSafeAreaInsets
uiManagerWillPerform
uiManagerDidPerform
uiManagerWillPerform
uiManagerDidPerform
..... many times
invalidateSafeAreaInsets (stops here)
```
=> I can toggle the flag once the event is emitted in `uiManagerDidPerform`, but the last event wont be sent

Gregoirevda:
@janicduplessis updated with clang format

janicduplessis (approved):

PR #623, opened 2025-04-30
Add IME Padding on Android

sookcha:
## Summary
- IME padding should be added when edge-to-edge is enabled as noted [here](https://stackoverflow.com/questions/79177674/windowsoftinputmode-adjustresize-doesnt-work-in-target-sdk-35/79364446#79364446)
- The keyboard avoidance will not work if the 'android:windowSoftInputMode' is set to 'adjustResize' and there is no IME padding.

## Test Plan

- Build app from Android 35 and below

### DIFF
--- a/android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaUtils.kt
+++ b/android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaUtils.kt
@@ -16,7 +16,8 @@ private fun getRootWindowInsetsCompatR(rootView: View): EdgeInsets? {
           WindowInsets.Type.statusBars() or
               WindowInsets.Type.displayCutout() or
               WindowInsets.Type.navigationBars() or
-              WindowInsets.Type.captionBar())
+              WindowInsets.Type.captionBar() or
+              WindowInsets.Type.ime())
           ?: return null
   return EdgeInsets(
       top = insets.top.toFloat(),


sookcha:
I'll open this after some testing

zoontek:
@sookcha Hi! I don't think this is a good idea. I had something similar in previous versions of react-native-edge-to-edge (when it was called react-native-bars), and it only interferes with [modern Android keyboard management](https://developer.android.com/develop/ui/views/layout/sw-keyboard) (= breaks `react-native-keyboard-controller`).

jacobp100:
Perhaps this could be exposed as an option on the SafeAreaProvider. Perhaps `insetsAndroid=['status-bar', 'display-cutout' ...]`

sookcha:
I didn't notice the conflict with `react-native-keyboard-controller` because I don't use that. I have a mixed feelings regarding `insetsAndroid=[...]` because I have a doubt introducing platform-specific param. But I think it might be a good idea considering we can choose which insets to include

zoontek:
@sookcha Modern keyboard APIs are available on Android since 11+.

We should really push users to use them (also, RN core KeyboardAvoidingView needs a serious update, that would help)

But mixing safe area insets and keyboard management like this is not the solution.

brentvatne:
@janicduplessis - @zoontek knows what he is talking about here so i think we can close this

sookcha:
Well. I still do think that in Android, keyboard insets are insets. I didn't quite understand what @zoontek means modern keyboard mgmt. Also, if it collides with `react-native-keyboard-controller`, then we can solve that problem separately.

PR #629, opened 2025-05-22
Prevent invalidateSafeAreaInsets get call when RNCSafeAreaProviderComponentView still in the fabric view pool

vikage:
…

Currently, RNCSafeAreaProviderComponentView is a fabric view. Fabric view using a reusable for views.
We already had `prepareForRecycle` to clean up the state of view when it was removed from the superview.
But we still have 2 things that can affect the state. 
The first one is the notification center. We have not removed it. Then `invalidateSafeAreaInsets` still get called when other places show the keyboard. It will lead to `_initialInsetsSent` being true, and our lib will not emit an event to js. (We have a return)

The second, UIKit still calls `invalidateSafeAreaInsets` when the keyboard shows. So we need a guard to prevent it affect our logic.

## Summary

We have a problem SafeAreaProvider empty children since `insets` is null, caused by other brownfield screen shows keyboard

### DIFF
--- a/ios/Fabric/RNCSafeAreaProviderComponentView.mm
+++ b/ios/Fabric/RNCSafeAreaProviderComponentView.mm
@@ -57,6 +57,7 @@ using namespace facebook::react;
 
 - (void)invalidateSafeAreaInsets
 {
+  if (self.superview == nil) { return; }
   // This gets called before the view size is set by react-native so
   // make sure to wait so we don't set wrong insets to JS.
   if (CGSizeEqualToSize(self.frame.size, CGSizeZero)) {
@@ -123,6 +124,7 @@ using namespace facebook::react;
   _currentSafeAreaInsets = UIEdgeInsetsZero;
   _currentFrame = CGRectZero;
   _initialInsetsSent = NO;
+  [NSNotificationCenter.defaultCenter removeObserver:self];
 }
 
 @end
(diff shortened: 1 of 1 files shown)


vikage:
@janicduplessis Please help to review. Thanks!

vikage:
@jacobp100 Please help me review also

ios/Fabric/RNCSafeAreaProviderComponentView.mm
```diff
@@ -123,6 +124,7 @@ - (void)prepareForRecycle
   _currentSafeAreaInsets = UIEdgeInsetsZero;
   _currentFrame = CGRectZero;
   _initialInsetsSent = NO;
+  [NSNotificationCenter.defaultCenter removeObserver:self];
```
jacobp100, line 127: Do we not need to re-add our listeners at some point?

vikage: @jacobp100 Oh right. I missing this part. I am adding this now
Thanks!

vikage: @jacobp100 I updated the code. Please take a look.

janicduplessis:
Thanks @vikage ! Can you just run the formatting script (yarn format:write)

vikage:
@janicduplessis The CI was passed. Please help me to merge it
Thanks!

jacobp100 (approved):

PR #642, opened 2025-07-01
fix building Fabric on macOS

gabrieldonadel:
## Summary

While refactoring `RNCSafeAreaProviderComponentView` in https://github.com/AppAndFlow/react-native-safe-area-context/pull/629 we ended up breaking builds for macOS because the `willMoveToSuperview` method does not exist in NSView's. 
Given that this function is just being used for registering notifications, which are not used in macOS, we can just add a platform check to the whole block without needing to use the equivalent `viewWillMoveToSuperview` function

## Test Plan

Before 

image

After 

image

image

### DIFF
--- a/ios/Fabric/RNCSafeAreaProviderComponentView.mm
+++ b/ios/Fabric/RNCSafeAreaProviderComponentView.mm
@@ -36,6 +36,7 @@ using namespace facebook::react;
   return self;
 }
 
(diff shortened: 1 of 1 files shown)


janicduplessis (approved):
Thanks!

Maybe I should look into running macos build on CI, seems pretty easy to break currently

PR #651, opened 2025-08-04
feat: Add Release build CI

mrousavy:
## Summary

This CI builds the Safe Area Context Example Android app everytime a new GitHub release is created.

The app will be built in **release mode**, and the resulting .apk will be uploaded and attached to the GitHub release.

We can also tweak the `on` trigger to always run it (e.g. for PRs too), but I don't think that makes sense.

This is super practical because users can just download the .apk from the releases page and maybe hunt down bugs more quickly instead of time travelling through git and rebuilding each time.

Also, this should probably reproduce the bug I was talking to Janic about in my Nitro CIs.

## Test Plan

Create a GitHub release and wait for this to build.

### DIFF
--- a/.github/workflows/android-release.yml
+++ b/.github/workflows/android-release.yml
@@ -35,7 +35,7 @@ jobs:
         working-directory: example
 
       - name: Build android example app with new arch disabled
-        run: ./gradlew assembleRelease -PnewArchEnabled=false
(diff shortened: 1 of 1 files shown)


janicduplessis:
Not sure why I haven't merged this, looks good!

janicduplessis (approved):

PR #653, opened 2025-08-07
feat: Set compile options using target_compile_reactnative_options

lukmccall:
## Summary

PR based on https://github.com/software-mansion/react-native-reanimated/pull/7776 by @tomekzaw.

To support `react-native@0.81.0`, we need to include `RN_SERIALIZABLE_STATE`. This can be done using `target_compile_reactnative_options`, which sets the necessary flags and definitions automatically, instead of setting them manually.

## Test Plan

- tested inside of expo/expo repo with RN 0.81 enable ✅


janicduplessis (approved):

janicduplessis:
@lukmccall thanks!

PR #658, opened 2025-08-22
chore(docs): update urls in documentation

samuelbeaulieu:
## Summary

This PR contains two changes:
- The GitHub URLs now point directly to the AppAndFlow repo.
- The tests URLs now point directly to the workflow files instead of using search filters.

## Test Plan

Nothing to test.


janicduplessis (approved):
Thanks for the fixes!

PR #670, opened 2025-10-14
fix: Add missing event dispatcher for tvOS and macOS in legacy arch

msynowski:
## Summary

https://github.com/AppAndFlow/react-native-safe-area-context/pull/592 has brought _eventDispatcher, but it's value is never set on tvOS and macOS as the assignment is inside a platform-specific `#if`

## Test Plan

Run an app with SafeAreaProvider on legacy arch. Insets should be available.


janicduplessis (approved):

janicduplessis:
Thanks!

PR #674, opened 2025-10-31
Fix WindowsInsets for apps targeting SDK 35+

volo-droid:
## Summary

This change fixes the following issue for apps targeting SDK 35+
https://github.com/facebook/react-native/issues/49759
 
Technical details:

> Before targeting SDK 35, using `android:windowSoftInputMode=”adjustResize”` was all you needed to maintain focus on — for example — an `EditText` in a `RecyclerView` when opening an IME. With `“adjustResize”`, the framework treated the IME as the system window, and the window’s root views were padded so content avoids the system window.
>
> After targeting SDK 35, you must also account for the IME using `ViewCompat.setOnApplyWindowInsetsListener` and `WindowInsetsCompat.Type.ime()` because the framework will not pad the window’s root views.

source: https://medium.com/androiddevelopers/insets-handling-tips-for-android-15s-edge-to-edge-enforcement-872774e8839b

## Test Plan

Reproducible project can be found here:
https://github.com/dprevost-LMI/rn81/tree/issue-497590-KeyboardAvoidingView-hide-input-under-SDK35

Adjust `SafeAreaView` style by removing `height` and changing `paddingBottom` to 32.

Without the fix `TextInput` is hidden whenever a soft keyboard is shown on SDK 35+, after the fix `TextInput` is correctly placed on top of the keyboard.

|Keyboard hidden|Current behavior|Fix applied|
|---|---|---|
|| ||

### DIFF
--- a/android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaUtils.kt
+++ b/android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaUtils.kt
@@ -9,6 +9,23 @@ import java.lang.IllegalArgumentException
 import kotlin.math.max
 import kotlin.math.min
 
+@RequiresApi(Build.VERSION_CODES.VANILLA_ICE_CREAM)
+private fun getRootWindowInsetsCompatV(rootView: View): EdgeInsets? {
+  val insets =
+      rootView.rootWindowInsets?.getInsets(
+          WindowInsets.Type.statusBars() or
+              WindowInsets.Type.displayCutout() or
+              WindowInsets.Type.navigationBars() or
+              WindowInsets.Type.captionBar() or
+              WindowInsets.Type.ime())
+          ?: return null
+  return EdgeInsets(
+      top = insets.top.toFloat(),
+      right = insets.right.toFloat(),
+      bottom = insets.bottom.toFloat(),
+      left = insets.left.toFloat())
+}
+
 @RequiresApi(Build.VERSION_CODES.R)
 private fun getRootWindowInsetsCompatR(rootView: View): EdgeInsets? {
   val insets =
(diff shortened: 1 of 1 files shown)


volo-droid:
Reproducible screen (taken from 
https://github.com/dprevost-LMI/rn81/tree/issue-497590-KeyboardAvoidingView-hide-input-under-SDK35 with adjustments to `SafeAreaView` bottom padding and height)

```javascript
import { NewAppScreen } from '@react-native/new-app-screen';
import { KeyboardAvoidingView, StatusBar, StyleSheet, TextInput, useColorScheme, View } from 'react-native';
import {
  SafeAreaProvider,
  SafeAreaView,
  useSafeAreaInsets,
} from 'react-native-safe-area-context';

function App() {
  const isDarkMode = useColorScheme() === 'dark';

  return (
    <SafeAreaProvider>
      <StatusBar barStyle={isDarkMode ? 'light-content' : 'dark-content'} />
      <AppContent />
      <SafeAreaView style={{ backgroundColor: isDarkMode ? '#000' : '#fff', paddingBottom:32 }} edges={['bottom']}>
        <KeyboardAvoidingView behavior={'height'} >
          <TextInput style={styles.textInput}/>
        </KeyboardAvoidingView>
      </SafeAreaView>            
    </SafeAreaProvider>
  );
}

function AppContent() {
  const safeAreaInsets = useSafeAreaInsets();

  return (
    <View style={styles.container}>
      <NewAppScreen
        templateFileName="App.tsx"
        safeAreaInsets={safeAreaInsets}
      />

    </View>
  );
}



const styles = StyleSheet.create({
  container: {
    flex: 1,
  },
  textInput: {
    flex: 1,
    marginHorizontal: 8 * 1.3,
    textAlignVertical: 'center',
    borderRadius: 21,
    backgroundColor: 'white',
    paddingHorizontal:  8 * 2,
    paddingVertical: 8,
    paddingTop: 8,
    minHeight: 8 * 5,
  },
});

export default App;
```

volo-droid:
Okay, I can see what the maintainers' concerns are. I've added a [comment](https://github.com/facebook/react-native/issues/49759#issuecomment-3492265009) to the original `KeyboardAvoidingView` issue summarizing the possible options to proceed with. I'll let the core RN team to handle it. Until then I assume we can keep this one open.

janicduplessis:
@volo-droid Thanks for working on this, but as I mentioned previously we decided to not include soft keyboard as part of insets on Android, this is mainly for keeping the behaviour the same as iOS where soft keyboard is not considered as a safe area inset. The recommended way is to also use https://github.com/kirillzyusko/react-native-keyboard-controller to handle the keyboard separately.

janicduplessis:
I added a note to the documentation about this https://github.com/AppAndFlow/react-native-safe-area-context/blob/main/docs/docs/usage.mdx#keyboard-handling

PR #686, opened 2026-01-12
Update broken React Native View prop links in API documentation

frubesss:
## Summary

These links were linking to the old docs. Updated them to point at the new docs.


janicduplessis:
Thanks!

PR #730, opened 2026-05-18
fix: remove UIImplementation usage from SafeAreaView

cortinico:
## Summary

I'm looking into removing `UIImplementation` from react-native core. 
That class is currently fully subbed and deprecated so calling it has no effect.

This library is still accessing it though. 

The `dispatchViewUpdates` call was already a no-op, so
removing it along with the `waitForReactLayout` helper that only
existed to synchronize with that dispatch.


janicduplessis (approved):
Confirmed this only affects old arch so fine with this.

cortinico:
> Confirmed this only affects old arch so fine with this.

Thanks for merging this @janicduplessis 🙏

PR #732, opened 2026-06-22
feat: upgrade to react-native 0.85.0 with RNTA 5.3.2

hurali97:
## Summary

> Note: Merge this PR before https://github.com/AppAndFlow/react-native-safe-area-context/pull/733

This PR is required for the AGP v9 adoption. With this PR, we upgrade to RN 0.85.0 with RNTA at 5.3.2.

## Test Plan

- Verified locally, app builds and launches fine 
- However, while selecting an option from dev menu, there is an infinite loop. I investigated a little but couldn't find root cause. Help here from maintainers is appreciated.

<details>
<summary>See Video</summary>

https://github.com/user-attachments/assets/8b64a241-8504-4bfb-93e5-6436aa967229

</details>


MeliValesca:
@hurali97 Thanks for the RN 0.85 upgrade!

I reproduced the Android restart loop. It was caused by a native Yoga crash in `RNCSafeAreaViewShadowNode::adjustLayoutWithState`, not by the dev menu itself.

The code was copying `props.yogaStyle`, modifying padding/margin on the copy, then replacing the whole Yoga node style. With RN 0.85/Yoga, that can hit a `StyleValuePool` assertion and abort the app. Since the selected example is persisted, the app restarts into the same crashing screen.

I pushed a fix that updates only the current Yoga node’s padding/margin values and marks it dirty when needed.

MeliValesca (approved):

hurali97:
Hey @MeliValesca 

Thanks for investigating and helping get this PR across the line. Appreciate it!

